PR #1, opened 2011-07-05
src.php

harikt:
Hi Paul ,

I have added the src.php , instance.php scripts.
Not sure why you missed it , if this is correct you can merge .

Thanks


harikt:
This will not work , for we need the Aura.Signal and Aura.Di . So closing it .

PR #12, opened 2012-04-19
i will use Aura.Sql in my new project

guweigang:
hi, i will use Aura.Sql in my new project.

Till now i met some issues, and have fixed them.
- READ.md: sample code, it seems dot(.) is higher then plus(+). 
- ConnectionManager: fix when config-array's key is 0 and ! 0 is 1 :-).
- add function Column::__set_stat() for var_exprot using.

There are some other problems here:
1. if do "insert/update/delete/..." on slave db-server, we should do something here to prevent it or switch to master db-server automatically instead.
2. if connect master for the first time, then till the end of this request, it will use this master-connection all the time.
3. for one request, there are two instance at most here, 1 for master and another for slave.

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,7 +269,7 @@ You can use profiling to see how well your queries are performing.
     
     // now get the profiler information
     foreach ($sql->getProfiler()->getProfiles() as $i => $profile) {
-        echo 'Query #' . $i + 1
+        echo 'Query #' . ($i + 1)
            . ' took ' . $profile->time . ' seconds.'
            . PHP_EOL;
     }
--- a/src/Aura/Sql/Column.php
+++ b/src/Aura/Sql/Column.php
@@ -143,4 +143,36 @@ class Column
     {
         return $this->$key;
     }
+
+    /**
+     *
+     * Returns column object for var_export. If you use "var_export" here, 
+     * there is another issue here. Saying that we're exporting an instance
+     * of the class 'foo\bar\BazClass'.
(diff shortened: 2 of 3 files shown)


harikt:
This is what I feel , if I understood correctly .

Qn : if do "insert/update/delete/..." on slave db-server, we should do something here to prevent it or switch to master db-server automatically instead.

If you are using connection manager https://github.com/auraphp/Aura.Sql/blob/master/src/Aura/Sql/ConnectionManager.php then getWrite() always get the master if there is master else to default connection. So insert , delete , update etc are not going to slave.

Qn : if connect master for the first time, then till the end of this request, it will use this master-connection all the time.

You can switch from one to the other with getMaster( 'name of connection' ) . Like the same getSlave( 'name of connection' )

Qn : for one request, there are two instance at most here, 1 for master and another for slave.

I didn't get what you meant here .

Thank you.

guweigang:
i mean that:
1. Sometimes, you need to invoke sql operations in multi-places (maybe different classes or methods). So there must be some global variables(static attribute also can do that, just like Singleton Pattern)  to retain sql-connections.

Here we got master and slave, you may want to switch between them. So why not retain two sql-connections? 
1. Sometimes, we don't know the type of the sql statement. If so, what can we do?

Just now, i check the type of sql-statement, if it's a type of write (not read), then connect to master automatically and all queries after it will use that master-connection.

what do you think?

PR #19, opened 2012-07-24
No such query type exception

harikt:
I was not able to run the phpunit . 
As travis also fails to run the unit test can you have a look into it ?
http://travis-ci.org/#!/harikt/Aura.Sql

or once I reach room I will check .

Thanks


harikt:
@pmjones I am closing this. There is no use with this for the autoloader throws the error . So we will not be getting in the try catch .

```
Class '\Aura\Sql\Query\Somestype' not found in
```

PR #20, opened 2012-07-25
env: values can be just strings

michaelklishin:
The `env:` key values can be just strings. They are then permutated and create a matrix, for each row of which you have the value `export`-ed. If you want to test against multiple databases, the `env:` technique is _just a convention_, your `before_script` lines need to do something with those env variable values.

At a minumum, you should not try to export values that will cause SSH sessions to wait on incomplete input. See [this doc guide](http://about.travis-ci.org/docs/user/languages/php/) and examples listed there (like Doctrine) to learn how this technique is used.


michaelklishin:
To make it clear: this approach is not only wrong and does not work the way you expect, it also causes stalled builds that slow down the entire php/perl/python queue for 30 minutes each. Please fix this ASAP.

michaelklishin:
I updated [our Database Setup guide](http://about.travis-ci.org/docs/user/database-setup/) to explain how it works better, here's a link to the [Doctrine2 .travis.yml](https://github.com/doctrine/doctrine2/blob/master/.travis.yml).

harikt:
Hi @michaelklishin  I am closing this issue for you have given PR #21 on develop which is the right way.

I am also interested to know whether you are from @travis-ci team.

I have added it and it was working fine as of http://about.travis-ci.org/docs/user/database-setup/  . Will check once again. If you are from @travis-ci team I don't want to wait . I can just merge ASAP .

Thanks

michaelklishin:
I am from the travis-ci.org team.

PR #28, opened 2012-09-28
Sorry the earlier PR didn't worked :( . Trying the other from the docs h...

harikt:
...ttp://about.travis-ci.org/docs/user/database-setup/

@pmjones sorry the earlier commit didn't worked as expected.  I have looked into the docs once again and made the changes. I hope this time it should work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,8 @@ env:
 php:
   - 5.4
 before_script:
-  - "mysql -e 'DROP DATABASE IF EXISTS test; create database test;'"
-  - "psql -c 'DROP DATABASE IF EXISTS test; create database test;' -U postgres"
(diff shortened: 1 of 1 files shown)


pmjones:
No worries -- applying now.

harikt:
This time it worked, but errors are different. I guess its due to the `tests/globals-dist.php` . Will check it.

harikt:
ok so the problem is with the globals-dist.php . Do you have any thoughts how we can make one for travis ? I wonder whether its a good idea to keep globals.php for travis ?

pmjones:
What _should_ the globals-dist.php values be? I'm happy if they change to match what Travis needs.

harikt:
If I am correct I am giving another PR #29 looking into the information from http://about.travis-ci.org/docs/user/database-setup/ 

Consider looking whether its correct :) .

harikt:
It fails :( . Wondering why the database not creating . I want to ask in travis support .

PR #32, opened 2012-12-01
Add setPdo, so people who already have a pdo can use the connection if t...

harikt:
This PR is from the conversation over twitter https://twitter.com/stanlemon/status/274708971217100800

@stanlemon Thank you for the gist  https://gist.github.com/4180449 .

I don't think we need to pass in the options a pdo object as in the gist. I feel we can introduce public method called `setPdo()` . So the people don't get confused of whether they need to pass a Pdo really .

I will add tests for this if @pmjones thinks its ok for a merge. I want to wait and hear what he thinks about this.

Hope that is ok with him and let me know if this solves your problem really.

```
use Aura\Sql\ConnectionFactory;
$connection_factory = new ConnectionFactory;
$connection = $connection_factory->newInstance(
     // connection name
    'mysql',

    // DSN elements for PDO; this can also be
    // an array of key-value pairs
    'host=127.0.0.1;dbname=aura',

    // username for the connection
    'username',

    // password for the connection
    'password'
);

$params = array(
    'driver'    => 'pdo_mysql',
    'user'      => 'username',
    'password'  => 'password',
    'dbname'    => 'aura'
);

$dsn = 'mysql:host=localhost;dbname=' . $params['dbname'];

$pdo = new Pdo($dsn, $params['user'], $params['password'], $options = array(
        PDO::MYSQL_ATTR_INIT_COMMAND => 'SET NAMES utf8',
    )
);

$connection->setPdo($pdo);
```

If you have any questions or concerns let me know.

### DIFF
--- a/src/Aura/Sql/Connection/AbstractConnection.php
+++ b/src/Aura/Sql/Connection/AbstractConnection.php
@@ -223,6 +223,18 @@ abstract class AbstractConnection
 
     /**
      * 
+     * Set the PDO connection object, normally you don't need to
+     * 
+     * @return void
+     *
+     */
+    public function setPdo(PDO $pdo)
+    {
+        $this->pdo = $pdo;
+    }
+
+    /**
+     * 
      * Connects to the database by creating the PDO object.
      * 
      * @return void
@@ -231,7 +243,9 @@ abstract class AbstractConnection
     public function connect()
     {
         $this->preConnect();
-        $this->pdo = $this->newPdo();
+        if (! $this->pdo) {
+            $this->pdo = $this->newPdo();
+        }
         $this->postConnect();
     }
 


stanlemon:
@harikt this looks fine to me - accomplishes the same thing.  Only potential downside to this is that a whole mess of values still need to be supplied at the factory to get an adapter out of Aura.Sql.  I was thinking about this earlier today, in order to use this in legacy code where I may want to phase in Aura.Sql and have an existing Pdo instance floating around I would have to either get some Db credentials or pass in dummy data, ala:
  $connection = $connection_factory->newInstance('mysq', 'foo', 'bar', 'baz');
  $connection->setPdo($pdo);

Which seems a little odd to me...  Not a big deal ultimately, and being able to inject a Pdo instance is better then not being able to, but part of me does wonder if something like...
  $connection = $connection_factory->newInstance($pdo);

Would have a more elegant method signature?

Just thinking out loud here, fwiw.

Thank you for putting together the PR btw.

harikt:
Hey @stanlemon , 

Thank you for writing down your feelings on this. Let me explain why I thought its a good idea to pass all the values.

1 ) There may be cases that your `PDO` instance is no more an instance, may be it lost the connection or died who knows. So we need to fetch the data anyway. In that case the `Aura.Sql` will be able to get the new `PDO` instance for you.

2 ) If you don't like to pass the real values keep the dummy ones also.

If you still feel there is a better way, please open the PR . I am happy to see the implementation and let's discuss.

stanlemon:
@harikt I'm not sure I understand #1 there - at least not how it would play out in practice.

To illustrate what I was describing... something like this:
https://github.com/stanlemon/Aura.Sql/tree/setpdo

If this isn't of interest though (totally fine) or not implemented in a way which you and Paul are comfortable with that's not a big deal.

Thanks for the consideration and the conversation - I appreciate it!

harikt:
So you were talking about the instance script where you would like to pass the pdo itself. Send it as another PR from it. We can discuss :+1:  .

harikt:
For a quicker view the commit https://github.com/stanlemon/Aura.Sql/commit/50a1d889b097fe1e25cf3c4952165a53dd3c87a2

harikt:
And I strongly feel we could not keep dsn as null ( https://github.com/stanlemon/Aura.Sql/commit/50a1d889b097fe1e25cf3c4952165a53dd3c87a2#L0R68 ) for this is a special case and we are making it possible for people to use a pdo if needed. And as there is a setPdo they can make use of this without any problem .

stanlemon:
Another options would be to substitue the Pdo instance in the dsn param, still specify the adapter name.  Really isn't necessary since the adapter name can be detected from the Pdo instance though.  It was that reason that caused me to add the null default so I would not an error on not supply that param.

pmjones:
Implemented in a modified version by:

https://github.com/auraphp/Aura.Sql/commit/d0e337bf2099d0bd860977de3256d2848a1d030c

PR #33, opened 2012-12-02
Add importing of a pdo instance to the Aura.Sql factory

stanlemon:
This stems from: https://github.com/auraphp/Aura.Sql/pull/32

This pull request includes the changeset from @harikt's setpdo branch as well.

The end goal is to use this like so:
<?php
$dsn =

$pdo = new \Pdo('mysql:host=localhost;dbname=test', 'root', 'root');

$connection_factory = include '../Aura.Sql/scripts/instance.php';
$connection = $connection_factory->newInstance($pdo);
$connection->connect();

$results = $connection->fetchAll('SELECT \* FROM Person');

### DIFF
--- a/src/Aura/Sql/ConnectionFactory.php
+++ b/src/Aura/Sql/ConnectionFactory.php
@@ -65,16 +65,23 @@ class ConnectionFactory
      */
     public function newInstance(
         $adapter,
-        $dsn,
+        $dsn = null,
         $username = null,
         $password = null,
         $options = []
     ) {
+        if ($adapter instanceof \Pdo) {
+            $pdo = $adapter;
+            $adapter = $pdo->getAttribute(\Pdo::ATTR_DRIVER_NAME);
+        } else {
+            $pdo = false;
+        }
+
         $class = $this->map[$adapter];
         $profiler = new Profiler;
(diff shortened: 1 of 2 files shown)


src/Aura/Sql/ConnectionFactory.php
```diff
@@ -65,16 +65,23 @@ public function __construct(array $map = [])
      */
     public function newInstance(
         $adapter,
-        $dsn,
+        $dsn = null,
```
harikt, line 68: And I strongly feel we could not keep dsn as null for this is a special case and we are making it possible for people to use a pdo if they already have. And as there is a setPdo they can make use of this without any problem .

stanlemon, line 68: Another options would be to substitue the Pdo instance in the dsn param, still specify the adapter name. Really isn't necessary since the adapter name can be detected from the Pdo instance though. It was that reason that caused me to add the null default so I would not an error on not supply that param.

pmjones:
Implemented in a modified version by:

https://github.com/auraphp/Aura.Sql/commit/d0e337bf2099d0bd860977de3256d2848a1d030c

PR #39, opened 2012-12-22
Added a whereIn and a orWhereIn method to the where traits

geissler:


pmjones:
I think this functionality is already supported.  You can do, for example ...

```
// $foo_list is an array of values
$select->where('foo IN (?)', $foo_list);
```

... or ...

```
// $foo_list is an array of values
$select->where('foo IN (:foo_list)');
$select->addBind(['foo_list' => $foo_list]);
```

... and the list arrays will be exploded, quoted, and comma-separated for you.  (This is something that happens at the DB adapter level.)

Is that existing functionality sufficient for your purposes?

pmjones:
@geissler Given my comment about existing functionality, is this still a problem?

geissler:
Sorry, i totally have forgotten that. The existing functionality is sufficient.

harikt:
@geissler cool. I am closing this @pmjones  :) .

Issue #40, opened 2013-05-04
Checking for property at the Column class


ghost:
When using Twig and trying to access a Column property, it gives an error saying that the method doesn't exists.

For example, if you try to access the column name property:

col.name

It halts saying the method name() doesn't exists.
It can be fixed by just putting an __isset($key) checking if $key is a class property.

Thanks!

pmjones:
Good call. Can you place a pull request so I don't forget?

ghost:
Ok. I'll do it tomorrow.
Thanks!

PR #41, opened 2013-05-06
#40 Adding the method __isset() to the Column class to fix the issue #40

ghost:
Pull request to fix the issue #40 (already closed by me)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 tests/tmp
 tests/globals.php
+/nbproject/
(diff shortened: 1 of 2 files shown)


pmjones:
My guess is that `property_exists($this, $key)` is more appropriate here.  Using `isset()` will return `false` if the property exists but is `null`.

ghost:
Yes, Paul, it's true... at the code I was developing I used property_exists and don't know why when I forked the project and created the pull I used isset().

Thanks for checking. Sending a pull with the correct one.

ghost:
Fixed.
If anything is incorrect, please, message me.
Thanks.

PR #48, opened 2013-08-18
added a proposed fetchEach method

nyamsprod:
this proposed method is complementary to the `fetchAll` method. It differs in the fact that it apply a callable function to each row.
Sorry if there are some mistakes in the `testFetchEach` method, I'm new to TDD.
Finally, the name is not set in stone.

### DIFF
--- a/src/Aura/Sql/Connection/AbstractConnection.php
+++ b/src/Aura/Sql/Connection/AbstractConnection.php
@@ -457,6 +457,35 @@ abstract class AbstractConnection
 
     /**
      * 
+     * Fetches a sequential array of all rows from the database and apply
+     * a callback to each found row
+     * 
+     * @param string|AbstractQuery $query The text of the SQL query; or, a
+     * query object.
+     * 
+     * @param array $bind An associative array of data to bind to the named
+     * placeholders.
+     * 
+     * @param callable $callback a callable applied to each row from the result set
+     * 
+     * 
+     * @return array
+     * 
+     */
+    public function fetchEach($query, array $data = [], callable $callback)
+    {
+        $stmt = $this->query($query, $data);
+        $res = [];
+        $index = 0;
+        while ($row = $stmt->fetch(PDO::FETCH_ASSOC)) {
+            $res[] = call_user_func_array($callable, [$row, $index]);
+            $index++;
+        }
+        return $res;
+    }   
+
+    /**
+     * 
      * Fetches a sequential array of all rows from the database; the rows
      * are represented as associative arrays.
      * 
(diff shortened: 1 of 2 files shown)


pmjones:
Neat idea! It's complementary to fetchAll(), but there's no corollary complement to fetchAssoc() (or the others that also return multiple rows). Perhaps it might be good to add a $callback param on all the fetch*() methods that return multiple rows?

nyamsprod:
Actually I have a local development with the callback parameter implemented directly on fetchAll, fetchAssoc, and fetchCol only since I don't see the added value on fetchOne and fetchValue. But as I'm new to unit testing I though first submitting this idea on a independent method to see if it was good and if my unit testing was also good. I thought about this after using AuraSQL where I had many returned rows with fetchAll and I had to make a second foreach on the resultset

pmjones:
That totally makes sense.  If we're going to do this, let's do it all at once on fetchAll/Assoc/Col and see what it looks like.  Can you resubmit with it applied to all of them?

As a side note, would it be enough to pass only the row to the callable?  I don't yet see why passing the index is needed (but it's not a big deal to me one way or the other).

Also, you _should_ be able to get away with $res = $callable($row) instead of call_user_func_array().

Thanks for working on this.

nyamsprod:
I have added the callback parameters as an optional parameter to the following methods:
- fetchAll
- fetchAssoc
- fetchCol
- fetchPairs

The callback can have up to 2 arguments, the current row and an optional index which will indicate the current iteration index. The index argument is taken from my use of the [JavaScript ForEach method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach)

Since I've made the changes I don't know if I have to make a new pull request or if it is ok like that ?

pmjones:
Tentatively accepting this; I reserve the right to reorganize as needed. Thanks!

pmjones:
Hi @nyamsprod -- what's the purpose of the $index being passed?  It seems redundant to passing just the array key.

nyamsprod:
I think the index key is better used in JavaScript and may have less value in PHP so if you wish you can remove it .. 
sometimes I use the $index to modify the callback behaviour let's say I want  action A to be taken on the first 5 iterations and another action taken latter on. I have to say that I never had to do this is PHP but I did it more that once in JavaScript. 
Other than that I don't think the $index is required, its a second optional argument so it can be skip by the 90% people who may use the callback parameters

pmjones:
/me nods

I'd like to remove the index entirely, then, if that's OK.  I get that it's useful in JS, but if you've never used it in PHP, let's leave it out until there's an actual need.  That sound good?

nyamsprod:
yes you can remove it

pmjones:
Excellent, thanks.

PR #50, opened 2013-09-04
query objects "where" method improvement

MAXakaWIZARD:
Change implicit "value" param to explicit in "where" method of WhereTrait, so it is visible to client code

### DIFF
--- a/README.md
+++ b/README.md
@@ -366,15 +366,24 @@ You can then modify the `Select` object and pass it to the `query()` or
 // create a new Select object
 $select = $connection->newSelect();
 
-// SELECT * FROM foo WHERE bar > :bar ORDER BY baz
+// SELECT * FROM foo WHERE bar > :bar AND zim = 'gir' ORDER BY baz
 $select->cols(['*'])
        ->from('foo')
        ->where('bar > :bar')
-       ->orderBy('baz');
+       ->where('zim = ?', 'gir')
+       ->orderBy(['baz']);
 
 $bind = ['bar' => '88'];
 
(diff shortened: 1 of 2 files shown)


pmjones:
This is actually a variation on something we did earlier, and then moved away from. The problem was that eventually a user will intend to pass "false" as a value to bind (sometimes it gets converted to an integer zero, sometimes to an SQL boolean representation).

MAXakaWIZARD:
Well, now we have "magic" param in this method now, I discovered it suddenly, when was looking in library's source code.

pmjones:
Yeah, it's a tough one. In a later variation we had an explicit param defaulting to a magic constant value (an MD5 hash of some string like "ThisValueIsReallyNotHere").  Instead of func_get_args() we checked against that magic constant value. That was weird too.

Maybe this is more a documentation issue?

MAXakaWIZARD:
What if we put there some object as default value?

pmjones:
Although a "null object" would be perfect here, PHP won't let you do `function ($foo = new Foo)`.

MAXakaWIZARD:
I think we have only two solutions:
1) Mention it in docs.
2) Create separate method with name "whereVal" or something like this and mention it in docs.

pmjones:
Of the two solutions, I like the first one better. If you feel like adding some notes on where() and having() methods, that would be really cool; if not, I will. Regardless, thanks for your attention to the details here.

pmjones:
(I'll close this PR when we have docs in place.)

harikt:
Closed via #51

PR #52, opened 2013-09-10
Sqlite-specific query objects

MAXakaWIZARD:
Also get rid of some duplication via creating LimitTrait, OffsetTrait and OrderByTrait


MAXakaWIZARD:
Link to SQLite SQL syntax documentation: http://www.sqlite.org/lang.html

harikt:
Nice and Thank you. It seems good to me.

harikt:
and we are about to release the package. @pmjones What do you feel on this?

harikt:
Just a different note @MAXakaWIZARD . Your site looks nice . Also do remember to checkout to new branch and commit, else you will get into trouble :-1:  .

MAXakaWIZARD:
Thanks! What do you mean "checkout to new branch and commit, else you will get into trouble". I regularly do pull upstream.

MAXakaWIZARD:
About release. I think we should wait until I implement PgSql query objects...

harikt:
@MAXakaWIZARD what I meant is when you commit to develop of your branch, and at some point the commit is not accepted, you may get into merge conflicts or the commits will be already there when you send a new PR. 

That is to get rid of these we make a new branch from develop and later send a PR on that merging with upstream.

I wrote something how I worked, may be there can be a better approach though, you can read at  http://auraphp.com/blog/2012/07/13/contributing-to-aura-project/

Thank you for the contributions.

harikt:
Do checkout subpackages of v2 (  https://github.com/auraphp/Aura.Sql/tree/subpackages ) . So we can have it in early stages itself.

Thanks

MAXakaWIZARD:
Ok, I see. Thanks for good points!

pmjones:
Sweet!

PR #53, opened 2013-09-12
PgSQL Query objects

MAXakaWIZARD:
Some refactoring and brand new PgSQL Query objects


pmjones:
Dude, you're a machine!

MAXakaWIZARD:
Well, it's a honor for me to work with you guys.

MAXakaWIZARD:
Planning to do the same for Microsoft SQL Server

pmjones:
Niiiice. Do you have access to an MS SQL server for testing? I do not, and so the MS SQL portions of the package may not work properly (although they used to).

MAXakaWIZARD:
I see. I think it's real to install MSSQL and run tests against it.

pmjones:
I recall it being more trouble than that; there's a PDO mssql driver that you need.

pmjones:
Maybe "pdo_sqlsrv" provided by Microsoft, and I think it's installable only on Windows.

MAXakaWIZARD:
Well, I use Windows, I'll try to do so, when will have some spare time.

pmjones:
Very good, sir.

harikt:
:+1:  good work @MAXakaWIZARD and nice to see your commits coming.

PR #54, opened 2013-09-12
Return specific query type from connection

MAXakaWIZARD:
e.g. Mysql connection will return Query\Mysql\Select instead of Query\Select and so on


pmjones:
There's only one thing wrong with this, and that is: there should be a way to get the "generic" query type from the connection. I envision a case where someone writes an app using only "generic" queries so that it is portable across MySQL, PostgreSQL, SQLite, etc.

MAXakaWIZARD:
What should I do? Create newGenericSelect method and so on?

pmjones:
There's two ways I can think of. Each has pros and cons.
1. Create a "use generic" property that gets set on the connection, maybe by constructor, maybe by setter method. Each newSelect() etc method checks that property and returns either a generic or a specific query object.  The problem here is that multiple classes may be using the connection, and some may want generic, and some may want specific.
2. Add a param to each newSelect() etc method to specify "I want a generic" (or a specific). The problem here is, do we default to generic, or to specific?

Your thoughts?

MAXakaWIZARD:
Maybe just add new method newGenericSelect along with newSelect?

pmjones:
The old behavior was the generic behavior; if there's a new method, it should be "newMysqlSelect" (etc).

MAXakaWIZARD:
Your're right

MAXakaWIZARD:
I think we should parametrize newSelect method. With default behaviour left untouched.

pmjones:
On consideration, I think I like the idea of adding the "newMysqlSelect()" (etc) methods. That makes it completely obvious what's going on, and there's no need to pass params every time you want something different.

MAXakaWIZARD:
Should we wait for Hari's thoughts?

pmjones:
@harikt should be awake in a little while; I have no problem hearing what he thinks here.

harikt:
I don't have a good answer here. I will say @pmjones can take the decision :+1: .

harikt:
looking at this I feel we are splitting the `newSelect` method to every where. And it seems not good to me for the current time.

May be we need to think how we can change in the factory `Aura\Sql\Query\Factory` ( For easy look up https://github.com/auraphp/Aura.Sql/blob/develop/src/Aura/Sql/Query/Factory.php#L37 )

harikt:
As `Select`, `Update`, `Delete` everything is moved to it's own class may be factory can get which connection and add it ? I need to look at it though.

pmjones:
With all that in mind, and in the interest of maintaining backwards compatibility, the new db-specific methods should be named `newMysqlSelect()` (etc). @MAXakaWIZARD let me know if you think there are real problems with that.

MAXakaWIZARD:
Ok then. I will do necessary changes

MAXakaWIZARD:
I'm done, please tell your opinion

pmjones:
My opinion is "merge it!" ;-)

For future reference, in your tests, you can use $this->assertInstanceOf('Class_Name', $object) instead of the get_class() stuff. I'll make the necessary changes on this one.

MAXakaWIZARD:
No-no. I did it via InstanceOf first. But it showed to me that it is not suitable for checking that object is instance of child class or something like that

MAXakaWIZARD:
Here's the example: http://3v4l.org/iD9bh

pmjones:
Ah so! Good call.

MAXakaWIZARD:
``` php
//imagine error here - it returns Query\Mysql\Select instead of Query\Select
$query = $connection->newSelect();
$isGenericQuery = $query instanceof \Aura\Sql\Query\Select; //true
```

pmjones:
Your point is understood, sir; no change will be made.

PR #56, opened 2013-09-30
Correct join order

MAXakaWIZARD:
This PR fixes wrong join order for Select queries:

``` php
$this->query
    ->from('t1')
    ->join('inner', 't2', 't2.id = t1.id')
    ->join('left', 't3', 't3.id = t2.id')
    ->from('t4')
    ->join('inner', 't5', 't5.id = t4.id');
```

Expected query is:

```
SELECT
            FROM
                "t1"
            INNER JOIN "t2" ON "t2"."id" = "t1"."id"
            LEFT JOIN "t3" ON "t3"."id" = "t2"."id",
                "t4"
            INNER JOIN "t5" ON "t5"."id" = "t4"."id"
```

Actual query is:

```
SELECT
            FROM
                "t1",
                "t4",
            INNER JOIN "t2" ON "t2"."id" = "t1"."id"
            LEFT JOIN "t3" ON "t3"."id" = "t2"."id"
            INNER JOIN "t5" ON "t5"."id" = "t4"."id"
```

### DIFF
--- a/src/Aura/Sql/Query/Select.php
+++ b/src/Aura/Sql/Query/Select.php
@@ -146,11 +146,6 @@ class Select extends AbstractQuery
             }
         }
 
-        // join these sources
-        /*foreach ($this->join as $join) {
-            $text .= $join . PHP_EOL;
-        }*/
-
         // where these conditions
         if ($this->where) {
             $text .= 'WHERE' . $this->indent($this->where);
--- a/src/Aura/Sql/Query/Select.php
+++ b/src/Aura/Sql/Query/Select.php
@@ -10,8 +10,6 @@
  */
 namespace Aura\Sql\Query;
 
-use Aura\Sql\Connection\AbstractConnection;
-
 /**
  *
  * An object for SELECT queries.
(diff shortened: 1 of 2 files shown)


pmjones:
Interesting! Does the FROM order matter in these kinds of cases?  That is, does having the FROM tables all in the same block give different results?

MAXakaWIZARD:
Well, from and join were implemented separately. First FROM, then JOIN. Actually this is wrong. Join should be connected to latest FROM in order of "join" method call, if any.

MAXakaWIZARD:
Today this thing bite me (I use Aura.Sql in one large project).

pmjones:
Wow, I've never run into that. Let me review and then I will pull.

If you can, it would be nice to get a variation of this into the develop-2 branch.  Since it was branched from develop, it might work, but it's a separate repo now, so it might not be as easy.  If you can't, I will.

As always, my many many thanks for your good work here.

pmjones:
Ah yes, I can see that you are right about connecting JOIN to each FROM. We'll probably want to change the indents too so that you can inspect the query "by hand" more easily.

MAXakaWIZARD:
Ok, I'll do that for v2 after merge.

MAXakaWIZARD:
Do you mean this repo: Aura.Sql_Query? Is it independent Query builders extracted from Aura.Sql? And we will proceed to develop v2 in that way?

pmjones:
Yes, that's the one.  It is in fact completely independent query builders; you can use plain old PDO with it if you like. And yes, I think v2 development will proceed in that way.

PR #68, opened 2014-03-21
Proxy PDO object and add support for injecting an existing PDO connection

stanlemon:
The only obvious thing I see this breaking is that ExtendedPDO->getDsn() will return null on existing. I shifted the tests around so that it would be easy to run them all on a newly constructed ExtendedPDO object, as well as one that has been injected into.


src/ExtendedPdo.php
```diff
@@ -124,26 +134,47 @@ class ExtendedPdo extends PDO implements ExtendedPdoInterface
...
+            // can't use array_merge, as it will renumber keys
+            foreach ((array) $attributes as $attribute => $value) {
+                $this->attributes[$attribute] = $value;
+            }
+        
+            // set the driver name
+            $pos = strpos($this->dsn, ':');
+            $this->driver = substr($this->dsn, 0, $pos);
+
+            $this->pdo = new PDO(
```
harikt: I don't think this is a right addition. You are extending `PDO` , then is there a need for creating a instance. Also this doesn't makes lazy connect.

jblotus:
I like the idea, but I would like to see this as almost a separate class instead. I think being able to reuse an existing PDO would be useful, but by design ExtendedPdo seems to want to be stand in as your main PDO instance.

stanlemon:
@jblotus The idea here should change the way that current ExtendedPDO is used, it can still stand in as your main PDO without any changes or penalties per se.

stanlemon:
@harikt Good catch man! My bad on that, the lazy connecting should be back to normal now.

pmjones:
@stanlemon I have merged this into a new branch, pdo-proxy. When the time comes I will merge _this_ one instead and apply the same changes so you get full credit via Github. Meanwhile, take a look at that one to see if you like the changes. In particular, do we really need a separate getDriver() method? Seems like we could co-opt the existing getAttribute() method on PDO itself.

PR #70, opened 2014-03-28
Use array_replace to merge attributes

stof:
I also casted options to array to ensure it is always an array because of the null default value, to avoid weird bugs (should it default to an empty array instead ?)


pmjones:
Not sure on `(array) $options` but I can check it here. Otherwise, very nice. Thanks!

PR #71, opened 2014-04-24
Fixing trailing commas and a documentation bug

jklein:
Trailing commas aren't allowed when calling a constructor, as is the code causes a parse error.  Also fixing a variable name to match the example above.

### DIFF
--- a/README.md
+++ b/README.md
@@ -356,7 +356,7 @@ $connections->setDefault(function () {
     return new ExtendedPdo(
         'mysql:host=default.db.localhost;dbname=database',
         'username',
-        'password',
+        'password'
     );
 });
 ?>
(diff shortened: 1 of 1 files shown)


harikt:
Thank you @jklein for your first contribution :+1:  .

harikt:
aah by the way are you using any of the aura packages at Etsy ?

jklein:
Not at the moment, but we might start using them in some of our open source projects via composer.  I'm currently using Aura for some side projects, and it's great!

harikt:
oh nice to hear @jklein and do join the google groups. There was some discussions lately on framework level stuffs.

harikt:
Have updated the contributors list and now you are also here http://auraphp.com/contributors/ :+1:  .

pmjones:
Thanks a ton @jklein !

PR #76, opened 2014-06-26
Correctly handle zero-indexed placeholder arrays on execute

### LINKED ISSUE #75: Sequential params on execution queries should be 0-indexed
Currently, fetch*(), perform(), etc. assume a 1-indexed array when dealing with non-named (question-mark) params. The base PDO execute() uses a 0-indexed array. This results in inconsistent behavior between ExtendedPDO and PDO, forcing workarounds like

https://gist.github.com/iansltx/6d0181c835e5afbf2965#file-gistfile1-php-L6

(I feed those results into fetchOne() or fetchAll() depending on whether I'm counting or retrieving results).

It sounds like the inconsistency here is because http://www.php.net//manual/en/pdostatement.bindvalue.php uses a 1-indexed array and http://www.php.net/manual/en/pdostatement.execute.php uses a 0-indexed array, when interacting with non-named params. php.net docs need an edit to clarify that this is what's going on on the execute site, but the editor isn't playing nice with me right now.

Fixing this issue will be backwards incompatible with the hack posted above, though for folks who've used 1-indexed as input instead and have a properly sized array it'd be possible to keep their implementations working.

iansltx:
No test yet; opening the PR to get feedback/test-running. Fixes #75.

### DIFF
--- a/src/ExtendedPdo.php
+++ b/src/ExtendedPdo.php
@@ -864,16 +864,16 @@ class ExtendedPdo extends PDO implements ExtendedPdoInterface
             return $this->prepare($statement);
         }
 
+        // match standard PDO execute() behavior of zero-indexed arrays
+        if (isset($values[0])) {
+            array_unshift($values, null);
+        }
+
         // rebuild the statement and values
         list($statement, $values) = $this->rebuild($statement, $values);
 
         // prepare the statement
         $sth = $this->prepare($statement);
-
-        // match standard PDO execute() behavior of zero-indexed arrays
-        if (isset($values[0])) {
-            array_unshift($values, null);
-        }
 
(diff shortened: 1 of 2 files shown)


iansltx:
Let me know if y'all want me to close this PR and re-open with just the second commit; the first one doesn't solve the problem.

harikt:
@iansltx I am seeing test also. or am I missing something?

This looks good to me. Paul will verify and merge. Be patient :+1:  .

Thank you.

iansltx:
When I made the PR originally I only had the first commit, which didn't include a test. The test is there now :grin:

pmjones:
Thanks for the test. This was an unexpected behavior from PDO; glad to see you caught it.

PR #77, opened 2014-06-26
Add HHVM compatibility

### LINKED ISSUE #72: Add HHVM back as a build target
It looks like the issue that previously blocked its usage has been fixed:

https://github.com/facebook/hhvm/issues/1560

There might be other blockers, but it's worth trying again IMO.

iansltx:
If the class constructor arguments when fetching a class are provided, HHVM's behavior diverges from Zend. I've got an issue out for that. In the mean time, intentionally leaving off the parameter gets rid of the behavior difference, so everything actually works. Not a fan of adding what amounts to HHVM-specific checks, but it does get us another available runtime.

...which would resolve #72

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ php:
   - 5.3
   - 5.4
   - 5.5
+  - hhvm
 before_script:
   - cd tests
 script: phpunit
--- a/src/ExtendedPdo.php
+++ b/src/ExtendedPdo.php
@@ -416,6 +416,11 @@ class ExtendedPdo extends PDO implements ExtendedPdoInterface
         array $ctor_args = array()
     ) {
         $sth = $this->perform($statement, $values);
+
+        if (!count($ctor_args)) {
+            return $sth->fetchObject($class_name);
+        }
+
         return $sth->fetchObject($class_name, $ctor_args);
     }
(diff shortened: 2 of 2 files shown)


harikt:
@pmjones I don't know whether this sounds a right merge. Rather than a hack I wish let them fix the issue at HHVM side.

pmjones:
Eh, I'm OK with it as an interim solution. No BC breaks, and technically it might be "more proper" to be more careful about how we do the the meta-programming stuff.

iansltx:
Dangit, now I'm responsible for code coverage dropping below 100%. Need to add a couple tests to cover a nonzero constructor arguments count, here:

https://scrutinizer-ci.com/g/auraphp/Aura.Sql/inspections/0a56b7e9-14db-42b5-88c3-f42ff1346af9/code-structure/operation/Aura%5CSql%5CExtendedPdo%3A%3AfetchObject?expandCoverage=1

https://scrutinizer-ci.com/g/auraphp/Aura.Sql/inspections/0a56b7e9-14db-42b5-88c3-f42ff1346af9/code-structure/operation/Aura%5CSql%5CExtendedPdo%3A%3AfetchObjects?expandCoverage=1

PR #79, opened 2014-07-26
Reset all the stored profiles.

harikt:


src/Profiler.php
```diff
@@ -109,4 +109,16 @@ public function getProfiles()
...
+    /**
+     *
+     * Reset all the profiles
+     *
+     * @return null
+     *
+     */
+    public function resetProfiles()
+    {
+        $this->profiles = array();
```
harikt, line 122: may be we can make this as a fluent one ?

PR #82, opened 2014-08-14
Use explicit data type when binding.

agentile:
Using explicit data types helped me prevent errors when binding to a boolean column, without using explicit data types `false` can become `""` and have the query fail. This was tested in postgres. Optionally I could have cast to an integer when setting the bind value, but that shouldn't have to be the case.


pmjones:
What happens when you bind a numeric string (e.g., "42") to an integer column? Anything?

agentile:
@pmjones It would result in queries like `UPDATE foo SET bar = '9'` as opposed to `UPDATE foo SET bar = 9` , which, at least in postgres, works just fine either way.

pmjones:
Accepting, but going to modify a bit afterwards. Thanks!

PR #89, opened 2015-01-02
Adding a callable configuration object.

brandonsavage:
Use case: we want to serialize the configuration of the SQL object, but Closure is not serializable. This object will allow us to create a closure.

For example:

``` php
$di->params['Aura\Sql\ConnectionLocator']['default'] = $di->lazyNew('Aura\Sql\CallableConfig', ['dsn' => $dsn']);
```

In this case, the CallableConfig is only created if we actually load the ConnectionLocator; the ExtendedPdo object is only created (and the connection is only created) if we load the connection.

### DIFF
--- a/src/CallableConfig.php
+++ b/src/CallableConfig.php
@@ -1,0 +1,88 @@
+<?php
+/**
+ *
+ * This file is part of Aura for PHP.
+ *
+ * @package Aura.Sql
+ *
+ * @license http://opensource.org/licenses/bsd-license.php BSD
+ *
+ */
+namespace Aura\Sql;
+
+/**
+ * This file offers a callable configuration object, to avoid creating the PDO connection
+ * until we need it. Replaces a Closure, in case we ever want to serialize the configuration.
+ */
+class CallableConfig
+{
+    /**
+     * A DSN compatible with PDO.
+     *
+     * @var string
+     */
+    protected $dsn;
+
+    /**
+     * The database username.
+     *
+     * @var null|string
+     */
+    protected $username;
+
+    /**
+     * The database password.
+     *
+     * @var null|string
+     */
+    protected $password;
+
+    /**
+     * The PDO options array.
+     *
+     * @var array
+     */
+    protected $options = array();
+
+    /**
+     * The PDO attributes array.
+     *
+     * @var array
+     */
+    protected $attributes = array();
+
+    /**
+     * @param $dsn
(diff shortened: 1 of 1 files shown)


pmjones:
Sorry, I don't think I understand the use case. Can you provide a little more background on why you think this is needed?

ghost:
Anonymous functions cannot be serialized. Objects that are callable can be serialized. If we want to use Aura.Sql together with a serializable Aura.Di, we need to be able to serialize the configuration for the Aura.Sql ConnectionLocator, which we can't do without this class (or without using the SerializableClosure library, which is black magic voodoo).

pmjones:
Ah so! If you don't mind, then, I'll revisit this when I review your Aura.Di work.

ghost:
Fine by me. Another option would be to leave this to the end user, but I find that being a bit more complicated.

pmjones:
(/me nods)

There might also be some generalizable solution that we can apply across the board. Thanks for this.

harikt:
Seems something what I was asking a few days back  https://groups.google.com/d/msg/auraphp/myU1o4dkYsU/eu_GfO4qvtYJ . 

May be https://github.com/jeremeamia/super_closure is what you need to serialize and deserialize things. But may be performance wise it have some drawbacks though.

pmjones:
I'm looking at this again, and while I get that a Closure cannot be serialized, why is it that a Closure is being used in the first place? One should be able to use lazyNew() for the ConnectionLocator params; those are serializable because they are callable.  E.g.:

``` php
    $default = $di->lazyNew('Aura\Sql\ExtendedPdo', ...);
    $read = array(
        'slave1' => $di->lazyNew('Aura\Sql\ExtendedPdo', ...),
        'slave2' => $di->lazyNew('Aura\Sql\ExtendedPdo', ...),
        'slave3' => $di->lazyNew('Aura\Sql\ExtendedPdo', ...),
    );

    $write = array(
        'master' => $di->lazyNew('Aura\Sql\ExtendedPdo', ...),
    );

    $connections = new ConnectionLocator($default, $read, $write);
```

Or am I still missing something?

pmjones:
Yeah, on further review, this looks totally achievable by using `LazyNew` for the default/read/write values. The DI container will not expand them right away, the ConnectionLocator invokes them with `call_user_func()`, they are serializable. Let me know if that solution does not do the trick for you.

PR #91, opened 2015-02-02
Adds a close method to null out the connection

jacobemerick:
Allows the manual closing out a connection to avoid issues with
long running scripts and timeouts. Method still has to be called
and has limited usefulness compared to re-connect options.

### DIFF
--- a/src/ExtendedPdo.php
+++ b/src/ExtendedPdo.php
@@ -634,6 +634,19 @@ class ExtendedPdo extends PDO implements ExtendedPdoInterface
 
     /**
      *
+     * Closes the current connection by null-ing out the underlying PDO
+     * instance. Useful for long-running scripts to avoid connection timeouts.
+     *
+     * @return null
+     *
+     */
+    public function close()
+    {
+        $this->pdo = null;
+    }
+
+    /**
+     *
      * Returns the profiler object.
      *
      * @return ProfilerInterface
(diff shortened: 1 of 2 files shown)


jacobemerick:
Thanks!

harikt:
Isn't this a BC break ?

jacobemerick:
@harikt not that I can tell. The additional method is a new feature unreferenced anywhere in the current code. It is a change in the interface, which may need to be documented, but I'm unsure how such things are handled.

harikt:
Here is another PR https://github.com/auraphp/Aura.SqlQuery/pull/41 which is not yet merged due to same reason :) .

harikt:
You could add a method in the ExtendedPdo but I don't think we can change an interface :-/ . What will happen if someone have used the interface ? If not no issues, else :P .

jacobemerick:
Oh! Got it. Hm... It's already merged in and there appears to be another very similar method added in PR/93, so I'm not sure what the thought is for the next release.

pmjones:
Rats, Hari is right about the interface. The SemVer-approved fix is to make another minor release, with the new method removed only from the interface; the implementation one can stay.

harikt:
Thank you. :-) .

pmjones:
This has been corrected as of https://github.com/auraphp/Aura.Sql/releases/tag/2.4.0

jacobemerick:
Awesome, thanks for the heads up!

Issue #22, opened 2012-08-15
Multiple Insert or Batch Processing


harikt:
Hi Paul , 

I am in need to issue multi query . Is there a way we can make it ?

I was looking at http://stackoverflow.com/a/4559320/487878 

Do you have any suggestions how we can make it ?

harikt:
I did some experiments to make this happen. But the problem is we are using `ValuesTrait` which has values stored as a single dimension via `col` and `cols` method. So the array of values will never become multidimensional .
I will continue with more ways if I get some idea. 
But this is a required one , else if I have 1000 values to insert I feel its a bad idea.

harikt:
I was trying something like 

```
public function __toString()
{
    if ( count($this->values) ==  count($this->values, COUNT_RECURSIVE) ) {
        $values = '(' . $this->indentCsv(array_values($this->values)) . ')';
        $columns = array_keys($this->values);
    } else {
        // for first one we need key
        $columns = array_keys($this->values[0]);
        $multi = [];
        foreach ($this->values as $value) {
            $multi[] = '(' . $this->indentCsv(array_values($value)) . ')';
        }
        $values = implode(',', $multi);
    }

    $string = 'INSERT INTO ' . $this->table . ' ('
         . $this->indentCsv($columns)
         . ') VALUES '
         . $values;
    return $string;
}
```

in `Aura\Sql\Query\Insert` and 

```
public function multiCol(array $columns)
{
    $this->values = [];
    var_export($columns);
    foreach ($columns as $k => $cols) {
        $value = [];
        foreach ($cols as $y => $col) {
            if ($k == 0) {
                $c = $y;
            } else {
                $c = $k . 'c' .$y;
            }
            $key = $this->sql->quoteName($c);
            $value[$key] = ":{$c}";
        }
        $this->values[] = $value;
    }
}
```

in `Aura\Sql\Query\ValuesTrait`  and `Aura\Sql\Adapter\AbstractAdapter`

```
public function insert($table, array $cols)
{
    $insert = $this->newInsert();
    $insert->into($table);
    if ( count($cols) ==  count($cols, COUNT_RECURSIVE) ) {
        $insert->cols(array_keys($cols));
    } else {
        $insert->multiCol($cols);
    }
    $stmt = $this->query($insert, $cols);
    return $stmt->rowCount();
}
```

Now the problem is in prepare statement and also 4 tests fails .

```
public function prepare($text, array $data)
{
    // need the PDO object regardless
[...]

PR #92, opened 2015-02-02
Add insert multiple rows with prepared statements emulation

sasha-x:
Here is my solution for https://github.com/auraphp/Aura.Sql/issues/22
Using pattern: 
INSERT INTO tbl_name (a,b,c) VALUES (?,?,?), ...
and "Rebuilder" way.
All comments and negative sql examples are welcome.
Or may be reopen issue and discuss it there?

### DIFF
--- a/src/ExtendedPdo.php
+++ b/src/ExtendedPdo.php
@@ -945,6 +945,10 @@ class ExtendedPdo extends PDO implements ExtendedPdoInterface
             return $this->prepare($statement);
         }
 
+		// rebuild multiline insert|replace pattern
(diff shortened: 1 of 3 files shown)


sasha-x:
I am sorry. Previous commit was broken.

pmjones:
I think this would be more appropriate in the Aura.SqlQuery package, so that a multi-insert can be put together programmatically and tested well. Cf. https://github.com/auraphp/Aura.SqlQuery/issues/16

sasha-x:
It's up to you.

pmjones:
Yeah, it'd be better as part of the Query system.

Issue #96, opened 2015-03-03
First parameter cannot be null when using perform/rebuilder


frederikbosch:
If the first parameter of a query with numbered placeholders is null, you will get an undefined offset in Rebuilder.php on line 220. The standard PDO library allows the first parameter to be null.

The following code generates the error.

``` php
$pdo->perform('SELECT * FROM test WHERE column_one = ?', [null]);
```

pmjones:
Fixed by #97

PR #97, opened 2015-03-03
Fix for issue #96. Allow first bind value of query with numbered placeholders to be null

frederikbosch:
Fixes issue #96 provided with a test.

### DIFF
--- a/src/Rebuilder.php
+++ b/src/Rebuilder.php
@@ -92,7 +92,7 @@ class Rebuilder
     public function __invoke($statement, $values)
     {
         // match standard PDO execute() behavior of zero-indexed arrays
-        if (isset($values[0]) || array_key_exists(0, $values)) {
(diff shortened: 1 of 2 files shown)


mlaphp:
Good. I say we get rid of the isset() entirely and go only with array_key_exists().

pmjones:
(Who the hell is mlaphp? ;-)

I say get rid of the isset() entirely and use only array_key_exists().

frederikbosch:
@pmjones No problem to get rid of isset.

pmjones:
Good deal! I'll let the CI stuff run and then merge. Thanks!

frederikbosch:
@pmjones No problem, thank you for the wonderful Aura libraries!

ps. I would be happy with a new version, could you tag one? :)

pmjones:
2.4.1 is released :-)

PR #108, opened 2015-08-04
Issue 104

JLacoude:
Here is a first version of a solution to the first two problems reported in Issue 104. To handle the `?` operator as the problem is also found when using pure PDO would require to analyze the statement detail and replace the operator with the corresponding function depending on the type of the left and right arguments.

The Rebuilder has been modified as to not require a PDO instance: when a passed value is an array, instead of replacing its placeholder by a csv string of quoted values it creates a csv string of placeholders.

Instead of using regular expression to split the statement it goes through it character by character and when a special one is found it calls a function to handle it. The `Rebuilder->statementPartsHandlers` contains those callables with the character to find as key.

To create Rebuilder for a database vendor you have to implement the RebuilderInterface.


tests/RebuilderTest.php
```diff
@@ -0,0 +1,196 @@
...
+
+        $expect = $stm;
+        $this->assertSame($expect, $actual);
+    }
+
+    public function testUnclosedMultiLineComment()
+    {
+        $stm = "SELECT *
+                FROM pdotest
+                /* Don't bind :foo";
```
harikt, line 48: seems you missed the closing comment `*/` ?

harikt:
Thank you for your good work on this.

One thing I do fear is breaking changes + additional requirements. Eg : mb_string etc.

tests/RebuilderTest.php
JLacoude, line 48: It is what we want to test: that the Rebuilder does not choke on a statement with unclosed comment.

harikt, line 48: Thank you for clarifying.

frederikbosch:
@harikt Sounds like a reason to increase the version then: 3.0.

pmjones:
This is fantastic! But it does introduce BC breaks, which means it would have to go into a new 3.x branch.

I see you has a "PostgresRebuilder" -- does all the logic in there work with any database, or just Postgred?

pmjones:
@JLacoude Actually, let me look at this a little more deeply. Requiring `mb*()` may not be that big a deal; adding the interface is probably a break, but there might be ways to remove the new interface in the interest of maintaining BC.

@harikt Besides the mb-string addition, and the new interface, can you note any other potential BC breaks?

harikt:
> Besides the mb-string addition, and the new interface, can you note any other potential BC breaks?

Thank you for the good work on it. To me, this is actually a complex PR. The rebuilder functionality is a complex one and not easy to quickly grasp.

JLacoude:
@pmjones 

> I see you has a "PostgresRebuilder" -- does all the logic in there work with any database, or just Postgred?

The PostgresRebuilder decorates the basic Rebuilder to add handling for the _$tag$ a string $tag$_ of writing strings in PostgreSQL. The basic Rebuilder should work with all databases. The goal would be to have one rebuilder per driver as each one has its quirks. Maybe using Traits for the shared parts.

If the BC break is too much, I can drop this PR and make one to correct the 2 bugs from issue 104 which are easily corrected and drop the dependency on PDO from the Rebuilder. And do the mb_string job in another PR for a possible version 3.

JLacoude:
> The rebuilder functionality is a complex one and not easy to quickly grasp.

I think part of the problem may come from the parameter passing around all methods. Maybe using a context object would make it easier and less prone to errors.

pmjones:
@JLacoude I really like the idea of the new functionality and would prefer to get in in 2.x if at all possible. Let's leave the mb-string stuff in place for now, but if you can reduce the complexity (or at least make easier to grasp) via the Context object you suggest, that would really help with review.

Thanks a ton for your work here.

frederikbosch:
@JLacoude Since #109 was merged, this will conflict your PR. That PR throws an exception when there is a missing parameter when using the numbered placeholders (question mark).

JLacoude:
As the changes seem to feel hazardous, I added a LegacyRebuilder. It is the default one and mostly does what the old rebuilder did: use regexp to extract parts of the queries. I just kept the decoupling from PDO when dealing with arrays.

To let users chose the Rebuilder they want, I added a setRebuilder method to ExtendedPDO.

frederikbosch:
@JLacoude I think that is a nice solution!

src/ExtendedPdo.php
```diff
@@ -992,6 +992,23 @@ public function prepareWithValues($statement, array $values = array())
...
+     * @param RebuilderInterface $rebuilder
+     *
+     * @return Rebuilder
+     *
+     */
+    public function setRebuilder(RebuilderInterface $rebuilder)
+    {
+        $oldRebuilder = $this->rebuilder;
+        $this->rebuilder = $rebuilder;
+        return $oldRebuilder;
```
harikt, line 1009: I don't think you need to return the old builder. What is the purpose of returning it?

JLacoude, line 1009: It's so you can easily swap it for some operations without needing a getter.

``` php
<?php
class Example{
    public function test(ExtendedPdoInterface $db){
        $original = $db->setRebuilder(new MockRebuilder());
        /* some operations */
        $db->setRebuilder($original);
    }
}
```

pmjones:
@JLacoude I have started a 3.x branch for this package, so BC breaks are allowed. If you are still willing, rebase and resend a PR to that branch, and we can merge it. Thanks for your efforts, and your patience.

JLacoude:
@pmjones Still willing to. But as BC breaks are allowed, would allowing back the use of multiple queries at once be a good idea?

pmjones:
> would allowing back the use of multiple queries at once be a good idea?

Might as well, since it's a BC break. Thanks for your help!

pmjones:
@JLacoude Any progress on this? I know it's a lot of work; thanks for any effort you can still provide.

JLacoude:
@pmjones Work and other projects have been taking a lot of my free time this year so I could not progress on this like I would I appreciated. I'll try to get something around the end of September.

pmjones:
@JLacoude No problem, and thanks for any efforts you can provide.

pmjones:
@JLacoude please see #134 as a replacement for this PR; all credit to you has been retained. Let me know if it looks OK. Thanks for all your good work here!

pmjones:
This is obviated by #138 which has been merged.

PR #109, opened 2015-09-02
throw exception when size of numbered placeholders does not matching size of input parameters

frederikbosch:
Now a `Undefined offset: 3` error is generated. This replaces that with a PDOException. Maybe I should introduce a separate exception for this. First see what you think.

### DIFF
--- a/src/Exception/MissingParameter.php
+++ b/src/Exception/MissingParameter.php
@@ -1,0 +1,22 @@
+<?php
+/**
+ *
+ * This file is part of Aura for PHP.
+ *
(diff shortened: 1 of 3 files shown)


harikt:
nice :+1:  .

pmjones:
I agree with @harikt : nice.

@frederikbosch can you change the the exception to a package-specific one? Call it Aura\Sql\Exception\MissingParameter and make the message little more explicit, something like "Parameter $x is missing from the bound values" (or something like that).

frederikbosch:
Updated and squashed.

pmjones:
Nice. Thanks!

Issue #110, opened 2016-01-18
yield with fetch*() functions


pavarnos:
# Hi Paul. Interested in your opinion on the value of using generators to reduce the memory footprint / increase performance. http://php.net/manual/en/language.generators.syntax.php eg fetchAssoc()

``` php
        // ...
        while ($row = $sth->fetch(self::FETCH_ASSOC)) {
            $key = current($row);
            yield $key => call_user_func($callable, $row);
        }
```

iansltx:
I'm not @pmjones but I've thought about building something that implemented ExtendedPdoInterface but used async mysqli under the covers for result sets, which is somewhat in the same vein, and would likely use generators like you're doing here.

Catch of course is that we'd lose PHP 5.3 support, but seeing as how it's been EOL for ages, that's probably fine.

To clerify...I'm in favor of this for the next major release.

Sieg:
Hi. And if possible, when the release is planned ? :)

harikt:
Hey @Sieg ,

Major means 3.x . Currently it is not yet started for this package.

Sieg:
@harikt, yes,

> To clerify...I'm in favor of this for the next major release.

Im interested in next major release date. Is it known already? This year summer, winter? next year ? :) As i see by commits history, the development of component is going not too fast?

harikt:
@Sieg Sorry, I don't know more about the release dates / plans. And it will be hard to maintain different versions and dump old versions just for yield functionality. And we don't have many developers here to maintain all these versions. So time is a constraint. Anyway things will be here eventually as far as this is not a bug, but a feature.

Thank you.

Sieg:
@harikt, thanks for the answer! The feature could be huge performance improvement for this layer. If you want more interest in your product, you should definitely introduce the implementation of generator usage in here ;) as soon as possible! Have a nice day!

harikt:
@Sieg cool, thanks for your interest. If anyone is interested I can create a 3.x branch. So can send PR against 3.x.

I would also love and let @pmjones take care rest of the things.

Thank you

pavarnos:
Just thinking out loud: the functions we are talking about are quite complex because of the optional callback passed in as a parameter. If there is going to be a new major release, why not drop the callback parameter to simplify things, and recommend lstrojny/functional-php for post processing results.

[...]

PR #112, opened 2016-03-05
added yield*() functions parallel to fetch*() functions. removed callables

pavarnos:
towards #110.
The diff looks a little weird in places: i added the new yield() functions directly below the fetch() functions, but removing the callable parameter in the original fetch()es seems to screw up the diff.
Did not do yieldGroup() because thought that it may not make sense as rows may be processed out of order?
Used modern array notation [] instead of array(). Happy to go back and fix that if needed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: php
 php:
-  - 5.3
-  - 5.4
   - 5.5
   - 5.6
   - hhvm
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,1 +1,1 @@
-This release modifies the testing structure and updates other support files.
+(no releases yet)
--- a/LICENSE
+++ b/LICENSE
@@ -1,23 +1,21 @@
-Copyright (c) 2011-2015, Aura for PHP
-All rights reserved.
+The MIT License (MIT)
 
-Redistribution and use in source and binary forms, with or without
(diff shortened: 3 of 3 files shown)


pmjones:
Thanks for this!

> The diff looks a little weird in places

Tell you what -- for purposes of reviewing the diff, can you put all the new `yield*()` in a block at the end? We can rearrange the methods later.

pavarnos:
aagh. please excuse this noob getting confused with git. made a new pull request. whats the best way to clean up this mess?

pmjones:
Heh, no problem. Looking over it now. The beatings will commence only if absolutely necessary. ;-)

pmjones:
OK, we're moving along nicely! Some style issues remain. If you could:

- Change `[ ]` to `[]` in the "empty array declarations (no space between the brackets).

- I don't think we need to wrap the `while()` conditions in double parens; a single set of parens should do. I.e., change from `while (($row = $sth->fetch(self::FETCH_NUM)))` to `while ($row = $sth->fetch(self::FETCH_NUM))`.

- Finally, if you could change uses of `array $values = array()` (and all similar) to `array $values = []`, that would be nice. I think they were like that before, so it's not your fault.

You should be able to make these changes in your existing branch; once you push them, this PR will auto-update itself.

How's that sound?

pavarnos:
this is the most fun i've had programming for a long time...

PR #113, opened 2016-03-06
added yield*() functions parallel to fetch*() functions. removed call…

pavarnos:
as per previous for #110

### DIFF
--- a/src/ExtendedPdo.php
+++ b/src/ExtendedPdo.php
@@ -198,7 +198,7 @@ class ExtendedPdo extends PDO implements ExtendedPdoInterface
      *
      * @return null
      *
-     * @throws PDOException if the connection fails.
+     * @throws \PDOException if the connection fails.
      *
      */
     public function connect()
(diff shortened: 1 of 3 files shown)


pmjones:
(Commenting on the right PR this time.)

OK, we're moving along nicely! Some style issues remain. If you could:
- Change `[ ]` to `[]` in the "empty array" declarations (no space between the brackets).
- I don't think we need to wrap the `while()` conditions in double parens; a single set of parens should do. I.e., change from `while (($row = $sth->fetch(self::FETCH_NUM)))` to `while ($row = $sth->fetch(self::FETCH_NUM))`.
- Finally, change uses of `array $values = array()` (and all similar) to `array $values = []`. I think they were like that before, so it's not your fault.

You should be able to make these changes in your existing branch; once you push them, this PR will auto-update itself.

How's that sound?

pavarnos:
ok. did that.
fwiw: i "code to green" in phpStorm: the little box in the top right corner of the window goes green if there are no errors or warnings in the code. I have the settings cranked way up. Without the extra brackets the code goes yellow and I get a warning "assignment in condition"

screen shot 2016-03-07 at 11 50 23 am

pmjones:
Nice.

src/ExtendedPdo.php
```diff
@@ -624,6 +521,118 @@ public function fetchGroup(
...
+     * Yields rows from the database
+     *
+     * @param string $statement The SQL statement to prepare and execute.
+     *
+     * @param array $values Values to bind to the query.
+     *
+     * @return array
+     *
+     */
+    public function yieldAll($statement, array $values = []) {
```
harikt, line 533: Good to move `{` to next line. Oh nothing other than style change :) .

Issue #114, opened 2016-03-06
class complexity reduction ideas


pavarnos:
Don't want to over-step my welcome here. Just wanted to toss a random idea at you to see if you like it...

phpStorm / php mess detector warn that the class is quite complex.

![screen shot 2016-03-07 at 11 05 27 am](https://cloud.githubusercontent.com/assets/589595/13557479/1492263e-e456-11e5-9409-4f6f690e5ca5.png)

To some extent this is to be expected: it is the nature of this kind of component. But I wondered if a couple of simple refactorings might reduce that complexity and offer a slight performance boost. 
1. Split the polymorphic constructor into a separate class. Make the main ExtendedPdo class take only a proper PDO dsn as its only parameter. Create a LazyExtendedPdo class that extends ExtendedPdo and move the username, password, options, attributes member variables there. Override connect() and disconnect() appropriately. The injected member variable is no longer needed.
2. Split responsibility for profiling out to a separate class. To do this, create a DecoratingExtendedPdo class as a decorator that has 

``` php
/** ExtendedPdoInterface */
protected $wrapped;

public function __constructor(ExtendedPdoInterface $wrapped) {
    $this->wrapped = $wrapped;
}
```

and proxies all its method calls to the $wrapped class. Create a ProfiledExtendedPdo class that extends DecoratedExtendedPdo and wraps each method call appropriately eg

``` php
public function rollBack()  {
    $this->beginProfile(__FUNCTION__);
    $result = $this->wrapped->rollBack();     
    $this->endProfile();
    return $result;
}
```

Then ExtendedPdo does not need to know about the profiler.

What do you think? Happy to submit 2 separate pull requests if that is helpful...

pmjones:
Not overstepping your welcome at all. Yes, it's definitely a complex class, one of the most complex in Aura. A lot of that is because it extends PDO, rather than being "its own thing".

Let's get the yield() stuff in place first, then we can revisit this more thoroughly.

In the mean time:
- The separate classes and constructors for a wrapped PDO vs a lazy-instantiated PDO sounds like it has merit.
- I am considering dropping the profiling as it is, and replacing it with a PSR-3 logging implementation.

Thanks for the critique on this.

pavarnos:
lol. was going to suggest a PSR-3 logger but didn't want to overstep my welcome even further...

pmjones:
Great minds :-)

pavarnos:
[...]

PR #116, opened 2016-03-08
split lazy connections into LazyExtendedPdo

pavarnos:
for #114


harikt:
Hi @pavarnos ,

Why did you not extended the ExtendedPdo in LazyPdo ? I wonder whether the lazy pdo is moving in the way like in version 1 : https://github.com/auraphp/Aura.Sql/blob/56bf2f90667442bc0eb5f35c4e0072e4f530fc98/src/Aura/Sql/Connection/AbstractConnection.php#L267-L276 .

pavarnos:
@harikt because ExtendedPdo is already a pdo instance, which we hold as a private member variable of LazyPdo. 

If LazyPdo also derived from ExtendedPdo there would be _two_ pdo instances per LazyPdo: one is the "LazyPdo extends ExtendedPdo", and one is the private member variable which is only instantiated after connect().

Also because #114 also talks about profiling / logging. One way to implement that is as a decorator wrapping an ExtendedPdoInterface, which is what LazyPdo and ExtendedPdo both implement. So then you can choose what to log and compose as it suits you.

The LazyPdo was a first step refactoring to reduce the complexity of ExtendedPdo. Next step is to pull out the profiler. Not exactly sure what pmjones has in mind for that: will wait and see if he accepts this pull request first.

harikt:
@pavarnos Thank you for the explanation.

pmjones:
@pavarnos I think @harikt has it right; we want for LazyExtendedPdo to match a PDO typehint, so it can be a drop-in replacement when refactoring away from "plain old PDO." (I'll need to examine the PR further.) Either way, thanks for the work, a refactoring of some sort has been desperately needed.

pavarnos:
OK. Fair comment. I wasn't thinking clearly about this: was more focussed on having two PDOs for the price of one (grin).

Pull request updated.

If you can give me some idea of how you wanted the profiling / logging to work I'm happy to do that as part of this PR or another...
What else is on the wish list for a 3.0 release?

pmjones:
All righty, next round of critique:

Splitting up the single ExtendedPdo from 2.x into two classes in 3.x leaves us with what are essentially several duplicated methods between the two classes (only 1 line of difference in a lot of them). My bet is that you can get away with putting a couple of stub or empty methods in ExtendedPdo, that only LazyExtendedPdo will use; or, perhaps, override perform() in LazyExtendedPdo() to connect(), then call parent::perform(). In any case, the goal would be to consolidate as much as possible into a single class, and let LazyExtendedPdo override only the bits it needs.

Do you see what I'm getting at there?

Again, thanks for your work here.

pavarnos:
@pmjones sorry I do not understand. Which methods were you thinking of?

pmjones:
@pavarnos cf #118 for an example of what I mean. (Yay, duelling PRs! ;-)

pavarnos:
very helpful. This PR updated with your ideas stolen from #118 (grin) plus a few extras answering @harikt. See the notes there.

harikt:
Awesome :-) . Good work @pavarnos .

Looks good to me.

I leave further comments for @pmjones .

Thank you.

harikt:
Hey @pavarnos ,

I was going through connect method :  https://github.com/pavarnos/Aura.Sql/blob/21b41e088b0fee18909023b740b7c8ee8fb6bf5b/src/ExtendedPdo.php#L45 and it looks to me the way @pmjones did in https://github.com/auraphp/Aura.Sql/blob/f90ffacc0caccbae6bf6e8f3fffeab5cbe615b63/src/DecoratedPdo.php#L122-L130 is good.

Regarding the disconnection it looks your way is good for we cannot guarantee whether we can disconnect the pdo instance created else where https://github.com/auraphp/Aura.Sql/blob/f90ffacc0caccbae6bf6e8f3fffeab5cbe615b63/src/DecoratedPdo.php#L141 , but there is a question of whether the user should allow to set the connection to null inside the ExtendedPdo .

I don't know though what is the best way to deal this :-/ .

pmjones:
@pavarnos The new changes look good; this is definitely moving in the right direction. I have other thoughts on how to reduce the complexity, and at least one other use case (that of decorating an existing PDO connection) but I think this is a good-enough step for now.

PR #117, opened 2016-03-11
(mistake) Split Lazy Connections From Decorated PDO

pmjones:
This is an example of what I meant for @pavarnos re: #116 .

This splits out the pseudo-polymorphic constructor of ExtendedPdo so that a DecoratedPdo has all the "real" logic, along with a stub connect() method that does nothing. The ExtendedPdo now extends DecoratedPdo, and overrides a few methods to provide lazy connection.

This is still a bit clunky; e.g., the class names probably need to be changed somehow. But I think it gets the idea across.

Thoughts?


pmjones:
Submitted against wrong branch.

PR #118, opened 2016-03-11
Split Lazy Connections From Decorated PDO

pmjones:
This is an example of what I meant for @pavarnos re: #116 .

This splits out the pseudo-polymorphic constructor of ExtendedPdo so that a DecoratedPdo has all the "real" logic, along with a stub connect() method that does nothing. The ExtendedPdo now extends DecoratedPdo, and overrides a few methods to provide lazy connection.

This is still a bit clunky; e.g., the class names probably need to be changed somehow. But I think it gets the idea across.

Thoughts?


pavarnos:
I think I get where you are going...
So your new DecoratedPdo has most of the functionality of the old ExtendedPdo. It proxies everything to the inner pdo class.
Your ExtendedPdo is a lazy-connection subclass of DecoratedPdo with little in it except the connection stuff? (and setAttibutes, which maybe belongs in your DecoratedPdo?)

I know we are not supposed to worry too much about performance, that good design is supposed to come first. But I can't help thinking that proxying every fetch_() and yield_() through an inner private member is going to cost something.

#116 has the advantage that the ExtendedPdo is actually a real Pdo in a valid state. (A proxied one like the 116 LazyConnectionPdo or the #118 DecoratedPdo has the outer Pdo in some invalid state because it bypasses the parent constructor)
The 116 ExtendedPdo eliminates a proxied function call for each method. ExtendedPdo really is an extension of Pdo: it adds extra methods onto the parent class without proxying, everything is in a valid state etc.
It pushes the cost for extra features (like lazy connections, profiling) out to sub classes with proxies. 

I think we are substantially going in the same direction. The original proposal in #114 suggested a generic "DecoratedPdo extends ExtendedPdo" that proxied every method to its wrapped instance. Then you could override where needed to add lazy connections, profiling etc. I just didn't want to take such a big step all at once and frighten you off (i guess). Its a pretty major refactoring... a bit arrogant for an outsider like me with no history in the project to just waltz in and change everything...

What do you think about implementing DecoratedPdo with a __call() magic method that passes everything to the wrapped inner pdo? Would reduce the line count / repetition significantly. Code completion could be done with @method doc block statements. I don't normally like magic, but in this case it seems warranted...

And yeah, your stub connect() / disconnect() methods on the base class look good: instinctively they should be part of ExtendedPdoInterface so the profiler can hook them as well.

tl;dr: my vote is for #116 over #118 for performance reasons. ExtendedPdo is exactly what it says on the tin: an extension of a Pdo. Other sub classes (profiling, lazy connections, whatever else comes later) can have the performance penalty of proxying: people don't have to use them unless they need that behaviour.

thanks for taking the time to talk this through

harikt:
Hi,

I am probably dump for the PR diff is confusing sometime. I would like to ask a question here in case if things will be easy for you all to see what I am looking at 

Working with symfony event dispatcher there is a traceable event dispatcher.

So what it does is wrap the orginal EventDispatcher in a TraceableEventDispatcher. So my question is when it comes to profiling and logging can't we do something similar ?

So we don't have a flag or anything in the ExtendedPdo but can wrap / pass the ExtendedPdo to a different class like ExtendedPdoProfiler / ExtendedPdoLogger ?

One drawback is  ExtendedPdoProfiler / ExtendedPdoLogger  will contain all the methods of the ExtendedPdo itself.

Eg : 

``` php
$pdo = new ExtendedPdo();
$pdoProfiler = new ExtendedPdoProfiler($pdo);
```

pavarnos:
@harikt yeah, just trying that out at the moment. So much repeated code: both the LazyPdo and the ProfiledPdo need to re-implement every method and pass it on to the proxied item. 
This would normally be fine in a smaller class, but the public interface is very large and it makes for a lot of repetition and potential bugs

 (magic methods didn't work: my misunderstanding)

pavarnos:
just committed refactored ProfiledExtendedPdo in #116. After a bit of experimentation with proxying (which was ugly), it seemed most natural to simply do "ProfiledExtendedPdo extends ExtendedPdo". It all seemed to fall together easily, after building on all the discussion above from you and @pmjones. 

Only weirdness was in the perform() method, where prepareWithValues() is also a profiled method and the profiling is not re-entrant.

Of note: scrutinizer gives us a score of 9.29 now, which is quite an improvement on the 6.29 of v2.x...

Issue #119, opened 2016-03-12
Remove "Lazy" ExtendedPdo in 3.x


pmjones:
It occurs to me that there are better ways to manage laziness of connections, than by having the connection manage itself. Providing a Factory or Container that only creates an ExtendedPdo on demand, rather than the LazyExtendedPdo instance creating its own internal connection on demand, would support a better separation of concerns. While this reduces the ability of older codebases to migrate directly to 3.x to get the benefits of lazy connection, the 2.x lazy connection remains available for that purpose; further, it streamlines and simplifies the ExtendedPdo logic itself.

@pavarnos @harikt and others: thoughts?

EDIT: Proof of concept branch in https://github.com/auraphp/Aura.Sql/tree/nonlazy .

harikt:
I am in the late nights, so I will keep this for a fresh day to think :) .

Thank you.

mbrevda:
This can cause issues when unit testing. I.e. if my class has an Sql instance passed in on construction, which tries to connect on jnstanciation, my test would fail if there is no DB available. But what if my test doesn't require a DB at All?

pmjones:
@mbrevda Fair point. For testing purposes, if a DB is not needed and not actually available, might it be worthwhile to pass a "FakeExtendedPdo" that overrides the constructor to not-connect at all?

mbrevda:
I think that is an oversimplification that went past the red line and is no back on over complicated territory. 

The test might be to confirm that the ExtendedPdo class is actually available - and yet we still require no connection.

pmjones:
(/me nods)

> The test might be to confirm that the ExtendedPdo class is actually available - and yet we still require no connection.

I presumed, but did not explicitly say, that FakeExtendedPdo would extend ExtendedPdo to override the constructor. Confirming that `$fakeExtendedPdo instanceof ExtendedPdo` should still pass in that case.

Keep the Devil's advocacy coming.

pavarnos:
what about making LazyExtendedPdo accept a callable in the constructor instead of dsn, username, password etc? This allows 
- laziness without having to use the connection locator (which is good for smaller projects), 
- allows decoration of an existing PDO (to cover that use-case which I completely overlooked in #116)
- simplifies LazyExtendedPdo: fewer private member variables
- would allow a mock PDO object to be injected for unit tests (where the mock is generated as needed by Mockery, phpUnit etc)

eg LazyPdoConnection:

``` php
[...]

Issue #104, opened 2015-06-10
Parameter extraction is wrong with some SQL syntax


JLacoude:
Using PostgreSQL at least three different cases of SQL keywords are considered parameters.

First is the cast operator "::"

`$pdo->perform('SELECT 1::TEXT', ['someparameter' => 'b']);`
This one can be avoided using the verbose way to cast but the short syntax looks better.
The problem comes from https://github.com/auraphp/Aura.Sql/blob/2.x/src/Rebuilder.php#L159 for which a quick fix could be `"/[^:]+(:[a-zA-Z_][a-zA-Z0-9_]*)|(\?)/m"`

Second one is using colons in comments:
`$pdo->perform('
        -- :should not be replaced
        SELECT 1', ['someparameter' => 'b']);`
`$pdo->perform("
        /* :should not be replaced */
        SELECT '2'", ['someparameter' => 'b']);`

The third problem comes with the new jsonb operators: ?, ?| and ?& 

`$pdo->perform('SELECT CAST(\'{"a":1, "b": 2, "c": 3}\' AS JSONB) ? \'b\'', ['someparameter' => 'b']);`

This third one can hardly be fixed by the user.

I think the best way to handle those cases and I'm sure a lot more with other RDBMS would be to not use regular expressions to parse the query but a simple state machine. And have one per driver.

pmjones:
Thanks for the report; extracting the params is a tricky business. I'm happy, but surprised, that there have been so few reports on this over the lifetime of that particular bit of code (10 years or so old by now).

> [do] not use regular expressions to parse the query but a simple state machine

/me nods

This sounds like a contribution candidate; if you can come up with one I'll review it.

pmjones:
This appears to be fixed by #138

PR #120, opened 2016-03-13
first attempt at Psr\Log profiling

pavarnos:
for #114


pavarnos:
travis build failed because i don;t know how to make it "composer install", and there is no composer.lock file in the original v2.x repo (by design?)

pmjones:
Accepting this as a provisional solution; further efforts might obviate it. Thanks for the work here. :-)

pavarnos:
Thanks. Came across a very tidy idea for lazy connections yesterday on stack overflow. Will make a pr in a couple of hours when I get to the office. 

What else is needed for a 3.0 release? (Apart from #104. Can I do anything useful there?)

pmjones:
> Came across a very tidy idea for lazy connections yesterday on stack overflow. Will make a pr in a couple of hours when I get to the office.

At your convenience.

> What else is needed for a 3.0 release?

#104, as you noted, which needs close attention.

A dev or alpha release of 3.0 could proceed even without that. Stable? Needs time in the trenches first.

pavarnos:
the idea for lazy connections didn't fit in with some of the other design goals mentioned in #119. for the record, the clever trick on SO was for LazyExtendedPdo to have no `private $pdo`, and 

``` php
function connect() {
    if ($this->is_connected) return;
    parent::__construct($this->dsn,$this->username,$this->password,$this->options);
    // ... set attributes etc
   $this->is_connected = true;
}
```

harikt:
@pavarnos I also mentioned the same on https://github.com/auraphp/Aura.Sql/issues/119#issuecomment-196166187 .

pavarnos:
@harikt: yes! sorry! you are correct. Maybe that is why it leaped out at me from SO: because I'd seen your idea here first.

harikt:
@pavarnos no problem and don't need to apologize.

pavarnos:
found myself writing this today... (developing against 3.x-dev)

``` php
class DatabaseConnection extends AbstractExtendedPdo
{
    public function __construct(\PDO $pdo)
    {
        $this->pdo = $pdo;
        // aura requires a profiler even if you don't want one
        $this->setProfiler(new Profiler(new NullLogger()));
    }
}
```

pmjones:
You should be able to extend DecoratedPdo in that case ... ?

pavarnos:
ah yes. of course.

PR #122, opened 2016-03-31
add callable factory to LazyExtendedPdo

pavarnos:
for #119


pmjones:
At this point, I don't think the factory is really necessary. (I reserve the right to revise my opinion later. ;-)

pavarnos:
agreed

PR #127, opened 2016-04-28
Adding PDO statement iterators (analog to "ExtendedPdo::fetch*" methods)

### LINKED ISSUE #125: Add statement iterator
Currently the `fetchAll` can be used to get all data from statement execution as an array. This unfortunately will result in out of memory error, when:
- there is a lot of data
- consumer of data needs only 1 row at a time to process it

Proposing to add several iterators, that would:
1. accept `PDOStatement` in constructor
2. while used will emulate output of `fetchAll`, `fetchAssoc` or even `fetchCol`

I've created them for a FOSS project of mine, but why not having something like this built-in in this great package.

Related Resources:
- http://qa.in-portal.org/diffusion/INP/browse/branches%252F5.2.x/in-portal/branches/5.2.x/core/kernel/db/db_connection.php;16347$1238-1519
- http://www.dragonbe.com/2015/07/speeding-up-database-calls-with-pdo-and.html

aik099:
In tests I'm comparing `fetch*` method result to corresponding `yield*` method result to confirm, that they work as expected.

Closes #125

### DIFF
--- a/src/ExtendedPdo.php
+++ b/src/ExtendedPdo.php
@@ -9,6 +9,11 @@
 namespace Aura\Sql;
 
 use Aura\Sql\Exception;
+use Aura\Sql\Iterator\AllStatementIterator;
+use Aura\Sql\Iterator\AssocStatementIterator;
+use Aura\Sql\Iterator\ColStatementIterator;
+use Aura\Sql\Iterator\ObjectsStatementIterator;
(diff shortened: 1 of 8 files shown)


pmjones:
Very nice. I'm sure to make some minor changes later, but the substance of the PR is good. Many thanks!

PR #128, opened 2016-05-05
Adding callable argument to yield* methods

nyamsprod:
Since Aura\SQL fetch\* methods can have an optional callable argument. I've added the same feature to the corresponding yield\* methods.
To do so:
- I've added a `MapIterator` class (inspired by the one I use on league/csv and downgraded it to work in PHP5.3) 
- I've updated the `PairsIterator` class to accept an optional callable

### DIFF
--- a/src/ExtendedPdo.php
+++ b/src/ExtendedPdo.php
@@ -1000,13 +1000,20 @@ class ExtendedPdo extends PDO implements ExtendedPdoInterface
      *
      * @param array $values Values to bind to the query.
      *
-     * @return Iterator\AllIterator
+     * @param callable $callable A callable to be applied to each of the rows
+     * to be returned.
+     *
+     * @return Iterator
      *
      */
-    public function yieldAll($statement, array $values = array())
+    public function yieldAll($statement, array $values = array(), $callable = null)
     {
         $sth = $this->perform($statement, $values);
-        return new Iterator\AllIterator($sth);
+        $iterator = new Iterator\AllIterator($sth);
+        if ($callable) {
+            return new Iterator\MapIterator($iterator, $callable);
+        }
+        return $iterator;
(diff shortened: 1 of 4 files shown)


aik099:
The signature of `yield*` method matches to one of `3.x` branch. There is no support for callable in there that's why I haven't added (when I was replicating these methods in `2.x` branch) such support in iterators either.

nyamsprod:
@aik099 I understand but this is one feature I use a lot when using Aura `2.x`, this makes the feature more inline with the `2.x` API.

aik099:
@nyamsprod , you mean callback feature? What are possible applications for these callbacks? I saw it's possible to use callback, but wasn't sure what benefit it provides compared to just doing `foreach` on result set.

nyamsprod:
The callable is used to simplify formatting the resultset by reducing the number of `foreach` used.

``` php
<?php

$resultSet = $pdo->yieldAll(
    'SELECT date_created FROM table WHERE 1',
    [],
    function(array $row) {
        $row['date_created'] = date_create_immutable($row['date_created']);
        return $row;
    }
);
```

aik099:
It does, but at cost of performance, when processing large data sets.

pmjones:
I have to agree with @aik099 here.

With fetchAll() etc, the idea for the callback is to process each result row as it gets created. The callback is needed because the creation process for each individual row is outside your control; it is hidden inside fetchAll() etc.

But with yieldAll() etc, the individual row creation process is back within your control. With yieldAll(), then, you would do this instead, given your above example:

``` php
$yield = $pdo->yieldAll('SELECT date_created FROM table WHERE 1');
foreach ($yield as $row) {
    $row['date_created'] = date_create_immutable($row['date_created']);
);
```

PR #134, opened 2016-12-12
Integrate @JLacoude 2.x rebuilder work into 3.x

pmjones:
This merges the 2.x changes from @JLacoude into the 3.x branch, resolving conflicts along the way.

I expect some further work will be needed, but this actually gets it into 3.x.

@JLacoude let me know if you see anything amiss!


frederikbosch:
@pmjones Since you are working on 3.x and thus there is no need for backward compatibility, I would suggest removing getters and setters (for profiler and rebuilder). There already is constructor injection.

frederikbosch:
Otherwise, great work!

JLacoude:
I currently have better rebuilders in a [branch](https://github.com/JLacoude/Aura.Sql/tree/issue-104-3) which handle more syntax specifics case from MySQL and Postgres. It also handle multiple queries but I never got to integrate it in the rest of Aura.

Most fetch* methods would have to be reworked or split to let users manage how they want to handle the results from a multiquery. For example do they want results from the last query only or an array of all result sets?

pmjones:
@JLacoude I think it's safe to skip the multi-query stuff for this iteration.

Do you want to set up a PR for your newer & better rebuilders? Or would it be better to merge this PR first, and then your newer work later?

JLacoude:
I'd like to set up a PR for those rebuilders if possible.

pmjones:
Yes, please do -- I am very eager to have them. :-)

pmjones:
@JLacoude any further progress on this? I admit I'd like to finish this sooner, rather than later.

JLacoude:
I got my branch in a state which can be merged with 3.x. But I don't know if it's normal to have almost no code coverage for DecoratedPdo, ExtendedPdo and Profiler.

If it is not a problem I'll make a PR.

pmjones:
Not a problem; after making the PR, all else can be dealt with. Thanks!

PR #137, opened 2016-12-27
hide password from stack traces, fixes #135

### LINKED ISSUE #135: [3.x] Do not distribute password in exports
I would suggest implementing `__debugInfo` and `sleep` to prevent passwords showing up in logging or debugging information. You can have a look at [Halite](https://github.com/paragonie/halite) how this could be implemented: see their `HiddingString` class with the [__debugInfo](https://github.com/paragonie/halite/blob/master/src/HiddenString.php#L66) and [__sleep](https://github.com/paragonie/halite/blob/master/src/HiddenString.php#L112).

If people agree on this, I am happy to create a PR.

frederikbosch:


src/ExtendedPdo.php
```diff
@@ -125,4 +125,24 @@ public function disconnect()
...
+     *
+     * @return array
+     *
+     */
+    public function __debugInfo()
+    {
+        return [
+            'args' => [
+                $this->args[0],
+                $this->args[1],
```
aik099: This is username, so I guess should be hidden as well.

frederikbosch: You are right.

tests/ExtendedPdoTest.php
```diff
@@ -651,4 +651,18 @@ public function testWithProfileLogging()
         $this->pdo->query("SELECT 5 FROM pdotest");
         $this->assertEquals(4, count($logger->getLog()));
     }
+
+    public function testNoExportOfLoginCredentials()
+    {
+        $pdo = new ExtendedPdo('sqlite::memory:', 'username', 'password');
+        $debugInfo = $pdo->__debugInfo();
```
aik099: Accoding to docs (see http://php.net/manual/en/language.oop5.magic.php#object.debuginfo) the `__debugInfo` method is called internally by `var_dump`. I think it would be more close to real-life scenario to call `var_dump($pdo)` and assert that printed fragments don't contain sensitive info.

frederikbosch:
@aik099 Try to test it, won't work within tests. Maybe with some sort of output buffering method, otherwise no.

tests/ExtendedPdoTest.php
aik099: Also maybe it would be better to see what properties are printed by `var_dump` prior to `__debugInfo` method being added and then return values from these properties (with credentials stipped). Currently you might be removing non-sensitive info, that was shown in `var_dump` result prior to this PR.

aik099:
> Try to test it, won't work within tests. Maybe with some sort of output buffering method, otherwise no.

@frederikbosch , maybe because it will only work on certain PHP version, where `var_dump` would call `__debugInfo` method internally? Since on Travis multiple PHP versions are tested then you need to adjust test method to be skipped on PHP versions that doesn't support `__debugInfo` call when `var_dump` is used.

OR

on non-support PHP versions assert that credentials aren't masked.

aik099:
@frederikbosch , ping.

frederikbosch:
@aik099 No, it is for different reasons. `var_dump` actually prints data and does not return anything. Hence, I cannot catch the output to test it. So then I thought, use `var_export`. But for `var_export` the magic call `__debugInfo` is not enabled. Hence, I cannot use `var_export` for testing either.

Remains two options:
1. Use `var_dump` in combination with output buffering (`ob_start` etc.)
2. Or test `__debugInfo` directly.

I choose the latter. Why? Unit tests are not there to test whether the programming language actually works. `__debugInfo` is documented by PHP. We do not need to verify that.

aik099:
> Hence, I cannot catch the output to test it.

You can actually:

```php
ob_start();
var_dump(...);
$data = ob_get_clean();
```

aik099:
> Unit tests are not there to test whether the programming language actually works. __debugInfo is documented by PHP. We do not need to verify that.

Incorrect. Currently if you make a mistake in `__debugInfo` method naming (e.g. `___debugInfo`) then tests calling it directly won't detect fact, that `var_dump` won't see it at all.

frederikbosch:
@aik099 These discussions are a bit tiresome, but ok, fine. There it is.

harikt:
@frederikbosch Thank you for the PR.

Thank you @aik099 for the review to make things better.

PR #138, opened 2017-01-11
Issue 104 3

JLacoude:
This PR should resolve multiple special cases when parsing MySQL or PostgreSQL queries.

The Rebuilder part can handle multiple queries separated by ";" but to expose it the ExtendedPdo class would have to be changed a lot.


pmjones:
Sweet. I'll look this over today/tomorrow and give feedback. Thanks!

pmjones:
Hi @JLacoude -- looking this over, it seems to have only Mysql and Postgres parsing. I am inclined to merge it as-is, then create a more basic parser that would be appropriate for SQLite, and then perhaps one for SQL Server. What are your thoughts on my doing so?

Cf. the new branch `3.x-parsers`.

JLacoude:
Go ahead, it is meant to be extended to have parsers for each SQL driver. I like how you reorganized things in the 3.x-parsers branch.

PR #140, opened 2017-01-19
Remove mb_ereg_* calls in 3.x

pmjones:
@pavarnos @jlacoude This PR removes the `mb_ereg_*()` calls in favor of `preg_*()` calls, with `/u` modifiers (for UTF-8 support). This is to move away from deprecated ereg-related functionality, and to remov possible state issues.

The tests are unchanged and continue to pass; however, I'd like a separate set of eyes to look it over.

Let me know what you think!

### DIFF
--- a/src/Parser/AbstractParser.php
+++ b/src/Parser/AbstractParser.php
@@ -218,8 +218,15 @@ abstract class AbstractParser implements ParserInterface
      */
     protected function handleSemiColon($state)
     {
-        $uselessCharacters = $state->capture(';\\s*');
-        $state->passString($uselessCharacters);
+        while (! $state->done())
+        {
+            $character = $state->getCurrentCharacter();
+            if (! in_array($character, array(';', "\r", "\n", "\t", " "), true))
+            {
+                break;
+            }
+            $state->passString($character);
+        }
         $state->setNewStatementCharacterFound(true);
     }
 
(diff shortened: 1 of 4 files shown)


JLacoude:
Well I wrote [some tests](https://gist.github.com/JLacoude/a0d1239df6ad123f4b9e2cf09f76522b) to check something and was surprised.

As expected preg_match with offset does not like multibyte characters. Even with the 'u' modifier.

Not really expected was the fact the multibyte regexp (mb_ereg_*) have the exact same problem.

The problem can be shown adding a test for the $sql = 'SELECT €$tag$ :foo $tag$'; string in PgSqlParserTest->testDollarQuotedStrings. If some string magician can add their two cents I'd be happy to learn new things. I'll investigate more during the week-end.

JLacoude:
After some experimentation I see two possiblities:
 - remove call to State->capture() and do things character by character. I implemented this and the only compromise is that the PgParser->handleDollar method accept characters which will be refused by Postgres.
 - making State->capture() work on a substring so there is no offset problem (the mb_substr function works perfectly). The problem is that by using preg_match instead of mb_ereg_* we are limited to UTF-8.

pmjones:
I say character-by-character. We're not validating the syntax or correctness of the query, just trying to replace array-based placeholders.

pmjones:
Using #141 instead. Thanks @JLacoude .

PR #141, opened 2017-01-22
Removing all calls to capture

JLacoude:
This way no call to any preg_match or mb_ereg_* function is necessary.


pmjones:
Dude. Sweet.

Issue #142, opened 2017-01-25
Performance issues with new parser


pavarnos:
New parser is quite a bit slower than the old one...
<img width="1473" alt="transactions_-_new_relic" src="https://cloud.githubusercontent.com/assets/589595/22279254/79521fe8-e32e-11e6-8997-b4f7997b4c46.png">
<img width="1031" alt="_new_relic" src="https://cloud.githubusercontent.com/assets/589595/22279256/7c83b794-e32e-11e6-9a61-74d8d6c61c07.png">

pmjones:
Yeesh, that's ugly. @jlacoude thoughts?

p.s. I figured it would be a noticeable difference, but not *that* noticeable. Maybe I'll put together a benchmark comparison. @pavarnos If you have a battery of query statements to parse, send them along.

pmjones:
It's pretty significant. Benchmark code:

```
$statement = "SELECT :foo, :bar, :baz";
$values = [
    'foo' => 'foo',
    'bar' => 'bar',
    'baz' => ['baz1', 'baz2', 'baz3'],
];

$k = 10000;
$begin = microtime(true);
for ($i = 0; $i < $k; $i ++) {
    $pdo->prepareWithValues($statement, $values);
}
$elapsed = microtime(true) - $begin;
echo "Total: " . ($elapsed) . PHP_EOL;
```

Results:
```
$ php bench-2.x.php;php bench-2.x.php;php bench-2.x.php;php bench-2.x.php;php bench-2.x.php
Total: 0.24196696281433
Total: 0.25062203407288
Total: 0.25587511062622
Total: 0.24196290969849
Total: 0.24846506118774
$ php bench-3.x.php;php bench-3.x.php;php bench-3.x.php;php bench-3.x.php;php bench-3.x.php
Total: 0.84113097190857
Total: 0.78416991233826
Total: 0.77567601203918
Total: 0.78816509246826
Total: 0.87434482574463
$ 
```

That's about a 3x difference on a very short query string; I imagine it gets worse as the string gets longer.

pmjones:
So, a series of questions:

- How important is it to automatically detect placeholders for array values, and expand them to comma-separated placeholders in the query string?

- If it's still important, how important is it to support non-utf-8 query strings?

- Is it possible to duplicate the character-by-character parsing wth a parser/rebuilder more like the older 2.x version, *and* avoid the bugs reported in #104 #107 #111 etc. ?

pmjones:
[...]

PR #143, opened 2017-02-01
simplistic rebuilder

pavarnos:
re #142 i wanted to try out a naive rebuilder that only replaced array parameter values.

Also added a CompositeRebuilder so it would be easy to make eg
- StripCommentsRebuilder
- QuoteIdentifierRebuilder
and other things people might think of later.

Caveats
- numbered parameters are not supported (though someone could add a NumberedParameterRebuilder that had a clever parser for those who needed it?)
- no whitespace around the `(:parameter_name)`

Just experimenting. Feel free to delete if it is way off track.


pavarnos:
overnight figured out how to use the existing parser to handle ? parameters if needed. Would be interested to see how this compares with the benchmark code from #142.

pavarnos:
added NullRebuilder: maybe useful for #142?

pmjones:
Closing in favor of #144. Thanks @pavarnos for working through this.

PR #144, opened 2017-02-07
Redo 3.x Parser to be more like 2.x Rebuilder

pmjones:
This PR converts the @jlacoude lexer parser logic to something like the 2.x rebuilder logic. It is not as full-featured as the lexer, but it is much faster. The bundled driver-specific parsers do not handle comments, but custom parsers can be injected for those cases.

As with the @jlacoude work, multiple named placeholders of the same name are allowed; likewise, numbered placeholders are converted to named placeholders. The renaming style has changed, so that repeated names are suffixed with `__` instead of `_`, and start at 1 instead of 0; likewise, the conversion of numbered placeholders to names now uses the name `__` instead of `__numbered`.

Tests for comments and multiple queries have been removed, as the new logic does not support them. Also removed is the PgsqlParserTest::testUnicodeStringConstants(), as I don't see how to tell the new parser logic to ignore colons uses as Unicode escapers in Postgres. Finally, the testEscapedCharactersInStringConstants() `SELECT "Escaping \\\\" :foo ""` is commented out, as it fails. Otherwise, all the pre-existing @jlacoude tests pass, though I have reorganized them under a base abstract test to remove duplication.

Finally, the benchmarks on this are much closer to the 2.x benchmarks than previous attempts.

$ php bench-2.x.php;php bench-2.x.php;php bench-2.x.php;php bench-2.x.php;php bench-2.x.php;
Total: 0.24769711494446
Total: 0.23618698120117
Total: 0.2351438999176
Total: 0.23142695426941
Total: 0.23610711097717
$ php bench-3.x.php;php bench-3.x.php;php bench-3.x.php;php bench-3.x.php;php bench-3.x.php;
Total: 0.25684905052185
Total: 0.26446294784546
Total: 0.25762104988098
Total: 0.27490496635437
Total: 0.27310609817505

This comes from the tradeoff of not handling comments and other cases.

@pavarnos @jlacoude @harikt please criticize as harshly as you can.


harikt:
@pmjones thank you for your good work. I know you have spent lot of time on this.

This have been in my thought for a while even though I have not been looking very closely.

1. I believe pdo can parse comments without any problem.
2. If pdo can parse comments Aura.Sql should be able to parse without much problem.
3. If Aura.Sql is becoming slow / cannot have the feature means it is a drawback.

So the question is what feature in Aura.Sql forces us to make the parser / lexer stuff ?

Can we have some way to get rid of this parsing and fix the issue that has bought us to go with the lexer / parser stuff ?

I noticed @pavarnos has also shared similar thought recently. I will look into the details of PR tomorrow.

Thank you

pmjones:
> Can we have some way to get rid of this parsing and fix the issue that has bought us to go with the lexer / parser stuff ?

The parsing is necessary for two reasons:

1. So we can bind array values to placeholders, which PDO does not allow. (I can go into more detail on this if you like.)

2. So we can specify more values to be bound than are actually in the statement; this allows you to pass (e.g.) an entire row of data to a query, and ignore the parts of the row that are not in the query.

Can't eliminate the parsing/rebuilding without losing those two features.

harikt:
> 1 . So we can bind array values to placeholders, which PDO does not allow. (I can go into more detail on this if you like.)

If I understand correctly you are mentioning about IN query. If so I probably like the idea of https://github.com/shadowhand/latitude#in-conditions . I haven't used it, but iirc Aura.SqlQuery do support the same. So something like a query builder can help to get rid of this right ? Or is my understanding wrong ?

> 2. So we can specify more values to be bound than are actually in the statement; this allows you to pass (e.g.) an entire row of data to a query, and ignore the parts of the row that are not in the query.

I understand it is really a pain when we see PDO giving you errors about bind params, but it may be a handy one for a developer to use the correct amount of bind values also.

> Can't eliminate the parsing/rebuilding without losing those two features.

Ya may be :-) .

I know one feature may be needed for one, another may be not needed for the other. As you always says tradeoffs :) . 

Have a nice day ahead.

pmjones:
@harikt If using a query builder, and all bound values are perfectly lined up, then you're right, no parsing/rebuilding is needed. In which case you can `setParser(new NullRebuilder())` (or pass it at construction time), which leaves the query string entirely alone.

PR #146, opened 2017-02-08
phpunit 6 and pds skelton to readme

harikt:


pmjones:
Hey @harikt thanks for this. 2 points:

- Leave pds/skeleton in require-dev (having it also in the README is fine)
- I think the setExpectedException() method calls are best left just before where the exception is expected, not as an annotation on the method as a whole

pmjones:
Hm, also, PHPUnit 6 requires PHP 7. I don't think we are bumping this library to PHP7 just yet (though for as long it has lingered, PHP 7 might be just fine).

harikt:
@pmjones if you are not going with php 6. It is ok, else `setExpectedException` is deprecated  https://thephp.cc/news/2017/02/migrating-to-phpunit-6 also see best practices : https://thephp.cc/news/2016/02/questioning-phpunit-best-practices .

harikt:
By the way if you are sure, you are not going with phpunit 6 and php 7, I will send the changes of README.md and composer.json as separate PR.

pmjones:
Yeah, so, replace setExpectedException() with expectException() then, instead of the annotation. In any case, that's only for PHPUnit 6 and PHP 7; for now, leave PHP 5.6 support in place, and the latest PHPUnit that supports it.

harikt:
Closing this PR for if phpunit is 5.7 there is no change needed.

https://github.com/auraphp/Aura.Sql/pull/147

PR #148, opened 2017-02-08
Change setExpectedException() with expectException()

harikt:


harikt:
I have tested this with 6, and 5.7 . To test for phpunit 6, find and replace `PHPUnit_Framework_TestCase` with `PHPUnit\Framework\TestCase` .

Issue #150, opened 2017-02-18
getPDO() or getInnerPDO()?


pavarnos:
Can we have a getPDO() or getInnerPDO() method?

Would be very handy when using phpDebugBar and other libraries which have their own PDO decorator.
I have code that builds the PDO and DecoratedPdo in a nice small tidy method that works for any environment. Later on, in another galaxy far far away, we check to see if we are on a dev server or in debug mode or whatever. What I want to be able to do is:

```php
$originalPDO = $container->get(DecoratedPdo::class)->getPDO();
$traceablePDO = new DebugBar\DataCollector\PDO\TraceablePDO($originalPDO);
$container->set(DecoratedPdo::class, new DecoratedPdo($traceablePDO));
```

which slides the debug bar in between the original pdo and the decorated pdo

pmjones:
Call it getPdo() and send a PR! :-)

pmjones:
As a side note, is `DebugBar\DataCollector\PDO\TraceablePDO` typehinted on `PDO`? Because `ExtendedPdo` extends PDO directly, so a PDO typehint should work there.

pavarnos:
Yes. see https://github.com/maximebf/php-debugbar/blob/master/src/DebugBar/DataCollector/PDO/TraceablePDO.php#L12

but the example code above was simplified quite a bit. I actually typehint against `class DatabaseConnection extends DecoratedPdo` because it makes stuff a lot clearer... long story... too boring to tell...

PR #151, opened 2017-02-18
add getPdo()

pavarnos:
ref #150

### DIFF
--- a/src/AbstractExtendedPdo.php
+++ b/src/AbstractExtendedPdo.php
@@ -426,6 +426,18 @@ abstract class AbstractExtendedPdo extends PDO implements ExtendedPdoInterface
 
     /**
      *
+     * Return the inner PDO (if any)
+     *
+     * @return \PDO
+     *
+     */
+    public function getPdo()
+    {
+        return $this->pdo;
+    }
+
+    /**
+     *
      * Returns the Profiler instance.
      *
      * @return ProfilerInterface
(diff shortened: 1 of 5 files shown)


pmjones:
A couple of things. First, ExtendedPdoInterface needs getPdo() on it. Second, getPdo() should always return a PDO instance; this means that DecoratedPdo can return it right away, but ExtendedPdo should connect() and then return.  Objections?

pavarnos:
ExtendedPdoInterface has getPdo() https://github.com/auraphp/Aura.Sql/pull/151/commits/812dfcf0e8771cb8ec3f94ef4383e8b2e8c205ed#diff-17e6536dc8cca90d75cc1d251abf5aa5R249
I've not been able to discern any pattern by which methods are grouped, so I added it after isConnected() in the class and interface.

re ExtendedPdo: so what should happen if the user calls disconnect() then getPdo()? Should it automatically reconnect? eg https://github.com/auraphp/Aura.Sql/pull/151/commits/812dfcf0e8771cb8ec3f94ef4383e8b2e8c205ed#diff-bdbc1b37686ca0aeb8da02735119783eR735

pmjones:
> ExtendedPdoInterface has getPdo()

My bad!

> I've not been able to discern any pattern by which methods are grouped

Alphabetical, IIRC.

> what should happen if the user calls disconnect() then getPdo()? Should it automatically reconnect?

I think that's the least astonishing thing to do.

pavarnos:
its not clear (to me) what went wrong with hhvm. any tips?

pmjones:
Not clear to me either. It seems to be recent, as of maybe 30 days ago, that HHVM started misbehaving.

pmjones:
@pavarnos Found the HHVM issue; the ExtendedPdoTest::testCall() method was blowing up. Resolved in a later commit.

PR #152, opened 2017-02-20
added phpunit via composer

syndicateSoftware:
phpunit tests passed with both version 5.7 and 4.8

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ php:
   - 5.6
   - hhvm
   - 7
+before_script:
+  - composer self-update
+  - composer install
 script:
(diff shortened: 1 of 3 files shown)


harikt:
Why does this contains have too much of commits? I see the message `don't need phpunit --version b/c version is output on run` repeated on all these PR's https://github.com/auraphp/Aura.SqlQuery/pull/128 , https://github.com/auraphp/Aura.Router/pull/145 .

Any reason you didn't do in a single commit ?

syndicateSoftware:
i just kept seeing things that i had forgotten to handle.

the commit is in both projects b/c i made the same mistake/fix in both

PR #155, opened 2017-03-14
Add link for upgrading from 2.x to 3.x

harikt:


CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 First alpha release.
 
+[See migration guide for details upgrading from 2.x to 3.x](https://github.com/auraphp/Aura.Sql/blob/695b240f1bd64dd5ba457935d7c0337be1055be5/docs/upgrade.md)
```
harikt, line 7: I have kept the specific commit for I don't know what will happen if we have more versions. Will the branches get deleted at that point. May be 3.x helps for now , so all the updated changes will also be reflected.

harikt:
:-) that was so fast... Never expected ;-) .

CHANGELOG.md
pmjones: That's fine; I think it goes well with that particular changelog entry. Having said that, a "final" conversion guide should be part of the 3.0.0 stable release.

pmjones:
Simple changes are easy to merge. :-)

CHANGELOG.md
harikt: > Having said that, a "final" conversion guide should be part of the 3.0.0 stable release.

Yes, that may be good. I have been looking how to reduce the changes mentioned regarding parser etc. For it is internal. So may be user level changes may be documented.

PR #161, opened 2017-06-22
Removed sqlsrv hack

### LINKED ISSUE #159: sqlsrv can't use errmode_exception
Long ago and far away (in 2015) there was a bug in pdo_sqlrv (see issue #90) which was fixed with a rather hacky detection of sqlsrv in the dsn string

However... it's 2017 and pdo_sqlsrv is long fixed (was fixed even before their github move) but the hacky workaround in code is breaking exception usage with the driver, you cannot set the error mode to exception using the extended pdo constructor

The hacky fix REALLY should be pulled back out 

I can happily do a PR for ripping the code back out ... right now I have a horrific copy and paste override constructor I'm maintaining just to remove the ick :(

Not sure what the best fix is though

auroraeosrose:
Removes a hack that completely blocks sqlsrv driver from using exceptions as an error mode.

The sqlsrv extension is now better maintained and this has been fixed for several versions.  https://github.com/Microsoft/msphpsql for the most recent.  If you're using a version old enough to still have issues, just set your own error mode in settings to something other than exception.. or upgrade the extension... really you should just upgrade your extension

Fixes #159

### DIFF
--- a/src/ExtendedPdo.php
+++ b/src/ExtendedPdo.php
@@ -65,13 +65,6 @@ class ExtendedPdo extends AbstractExtendedPdo
             $options[PDO::ATTR_ERRMODE] = PDO::ERRMODE_EXCEPTION;
         }
 
-        // sqlsrv fails to connect when the error mode uses exceptions
-        $sqlsrvWarnEx = substr($dsn, 0, 7) == 'sqlsrv:'
-            && $options[PDO::ATTR_ERRMODE] == PDO::ERRMODE_EXCEPTION;
-        if ($sqlsrvWarnEx) {
-            $options[PDO::ATTR_ERRMODE] = PDO::ERRMODE_WARNING;
-        }
-
         // retain the arguments for later
         $this->args = [
             $dsn,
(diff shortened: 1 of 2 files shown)


pmjones:
This is awesome. However, tests fail, probably because they're testing to see if sqlsrv sets to warnings instead of exceptions. If you have time, rip that test out (or otherwise correct it) and I can merge. Thanks for the PR!

pmjones:
KILL IT WITH FIRE

auroraeosrose:
Ok - tests are ACTUALLY passing now but hhvm is saying

HHVM is no longer supported on Ubuntu Precise. Please consider using Trusty with `dist: trusty`.

so should I file a bug for that :)

pmjones:
No need, I'll attend that at release time. Again, thanks!

pavarnos:
Some text in the readme.md also needs to be removed.

pmjones:
@pavarnos Gah, of course. If you or @auroraeosrose can PR that, so much the better; if not, I can handle it myself later on.

Issue #163, opened 2017-10-25
getAttribute() pass-through calls do not work


AlexHowansky:
Using 3.0.0-beta1:
```
$db = new \Aura\Sql\ExtendedPdo(...);
$driver = $db->getAttribute(\PDO::ATTR_DRIVER_NAME);
```
This returns nothing and emits the following warning:
```
PHP Warning:  PDO::getAttribute(): SQLSTATE[00000]: No error: PDO constructor was not called in ...
```
Forcing a connect with the `connect()` method before calling `getAttribute()` does not work. I can work around the issue by doing this:
```
$driver = $db->getPdo()->getAttribute(\PDO::ATTR_DRIVER_NAME);
```
However, I don't believe this should be required, as the `getAttribute()` method should still work correctly in the extended class.

harikt:
The problem why it is not working is mentioned on https://github.com/auraphp/Aura.Sql/pull/164#issuecomment-341367053 .

For the current time, as you can rely on the `getPdo` please make use of it.

Thank you.

harikt:
Merged. So closing the same.

PR #164, opened 2017-10-26
Fix issue #163

sunkan:
Add getAttribute and setAttribute functions to make sure pdo is
connected before calling the functions

### DIFF
--- a/src/AbstractExtendedPdo.php
+++ b/src/AbstractExtendedPdo.php
@@ -979,4 +979,31 @@ abstract class AbstractExtendedPdo extends PDO implements ExtendedPdoInterface
                 return;
         }
     }
+
+    /**
+     *
+     * Retrieve a database connection attribute
+     *
+     * @param int $attribute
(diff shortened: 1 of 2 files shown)


harikt:
Something that troubles me is : If `getAttribute` is not working, I think the `__call` will not work for other methods of pdo. That is a serious problem which need to be looked into I guess.

sunkan:
But the problem is that we extends PDO. 
And because of that `__call` will never be called for native pdo functions

harikt:
oh oh I see what you mean. 

Then the `__call` looks like a function that will never be called.

Thank you.

sunkan:
The only other function that isn't implemented from the base class is `getAvailableDrivers` and that's a static function.

So the only missing functions in `AbstractExtendedPdo` where `setAttribute` and `getAttribute`

harikt:
@pmjones do you have any thoughts on this, else I will merge this.

PR #170, opened 2018-04-23
Fixed error Class 'Aura\Sql\PDO' not found

JoeBengalen:
prophecy is giving me this error when using the `ExtendedPdoInterface`:

```
✘ Fatal error happened while executing the following
    xxxx
    Class 'Aura\Sql\PDO' not found in ...\vendor\phpspec\prophecy\src\Prophecy\Doubler\Generator\ClassMirror.php on line 214
```

This is because `fetchGroup` is using a `PDO` constant.


pmjones:
Wow, I can't believe we never ran into that before. Thanks!

Issue #175, opened 2018-06-13
Attempt to save boolean values results in silent failure


kevinsmith:
Using the `perform()` method to prepare and bind a boolean value for an insert results in a silent failure. The record is not inserted, and we only discovered it because Atlas threw an exception— `Expected 1 row affected, actual 0.`. I was able to recreate the bug without Atlas.

Recreate the bug with this table:

```
CREATE TABLE `test_table` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `name` varchar(30) NOT NULL,
  `active` tinyint(1) NOT NULL DEFAULT '1',
  `default` tinyint(1) NOT NULL DEFAULT '0',
  PRIMARY KEY (`id`)
) ENGINE=InnoDB AUTO_INCREMENT=1 DEFAULT CHARSET=utf8;
```

And this `ExtendedPdo` call:

```
$pdo->perform(
  'INSERT INTO `test_table` ( `name`, `active`, `default` ) VALUES ( :name, :active, :default )',
  [
    'name'    => 'Item Name',
    'active'  => true,
    'default' => false,
  ]
);
```

Result: No errors or exceptions, but the insert is not successful.

Aura.Sql 2.5.2
PHP 7.1.17
MySQL 5.7.17.

To be fair, this is actually an almost 12 year old (!!!) [open bug in PDO](https://bugs.php.net/bug.php?id=38546), but since Aura.Sql is an extension of the native PDO implementation, there's an opportunity here to easily provide a fix.

Our workaround was to cast the boolean values as integers before making the `perform()` call. It would seem something similar could be done in [bindValue()](https://github.com/auraphp/Aura.Sql/blob/2.x/src/ExtendedPdo.php#L1118).

pmjones:
Great catch. Send a PR if you can, otherwise I will attend as time permits. Thanks!

kevinsmith:
It occurs to me that this change may have unintended consequences for PDO instances using drivers other than MySQL. I'm unfamiliar with most of them, so I can't imagine how they might behave when expecting a boolean and receiving an integer. What are your thoughts on including driver-specific behavior in this library?

RyanCasas:
@kevinsmith, could you please test if #176 solves your issue? If it does so, we could write a phpunit test for this use case.

hakuno:
Hey guys @pmjones @kevinsmith look at PR sent by @RyanCasas :+1:

harikt:
[...]

PR #176, opened 2018-06-26
This should solve issue #175, and it is MySQL only.

RyanCasas:
It delegates the binding of the integer to the `ExtendedPdo` interface, this way we have no duplicated code.

### DIFF
--- a/src/AbstractExtendedPdo.php
+++ b/src/AbstractExtendedPdo.php
@@ -912,6 +912,10 @@ abstract class AbstractExtendedPdo extends PDO implements ExtendedPdoInterface
         }
 
         if (is_bool($val)) {
+            if($parser instanceof Parser\MysqlParser) {
+                return $this->bindValue($sth, $key, $val?1:0);
+            }
+
             return $sth->bindValue($key, $val, self::PARAM_BOOL);
         }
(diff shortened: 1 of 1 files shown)


src/AbstractExtendedPdo.php
```diff
@@ -912,6 +912,10 @@ protected function bindValue(PDOStatement $sth, $key, $val)
         }
 
         if (is_bool($val)) {
+            if($this->parser instanceof Parser\MysqlParser) {
```
hakuno: Add whitespace `if ($this->parser instanceof ...` on it. For PSR's sake.

RyanCasas: Whoops! Thanks for the revision.

pmjones:
@RyanCasas I ended up putting a separate fix in place, https://github.com/auraphp/Aura.Sql/commit/d1283359bb552571a79568e91e64bc7f2dc4f8ac . 

The other solution is slightly different; my testing indicated that PDO wants a string "1" or "0", not merely integers.

On consideration, I should have merged this first and then edited it; my apologies for not doing so. Thank you for putting forth the effort on a fix in the first place.

PR #180, opened 2019-05-17
remove unused param in queryparser

r4ndsen:


harikt:
Apart from hhvm issue on travis rest looks good. Not sure why there was `$sub` used and whether it is missed. So requesting @pmjones  for review.

r4ndsen:
> Apart from hhvm issue on travis rest looks good. Not sure why there was `$sub` used and whether it is missed. So requesting @pmjones for review.

that was what i looked up first, but the param isnt used in either of the child classes that are derived from that.
i found this from using [rector's](https://github.com/rectorphp/rector) dead-code analysis

harikt:
Sorry for being late. Thank you.

PR #188, opened 2021-02-12
Support PHP 8 on Aura.Sql 2.x

sudo-plz:
This release adds PHP 8 support until we migrate to the 3.x version.

Minimum required PHP version has been increased from 5.3 to 5.6 to make it possible.


pachico:
Is there any plan to accept this PR?

pmjones:
Maintaining backwards-compatibility with 5.3 is a non-negotiable element for this library. With that in mind,  I'm inclined to accept the PR on the .travis.yml file as-is, and on the composer.json file only if PHP 5.3 remains as an allowed version. 

And a question: is it really the case that this package will not operate on PHP 7 and 8 without the changes to ExtendedPdo::query() and PdoInterface::query()? The spread operator is great, but is the change strictly necessary for operation on PHP 7 and 8? (I have not tried so I cannot say for sure, but my intuition is that those changes are not strictly necessary.)

sudo-plz:
Hi Paul,
As far as I know it is not possible in the 2.x repo to support PHP 5.3 and 8.0 at the same time.
Here is an [example](https://3v4l.org/AWp4F) to demonstrate the error we're facing.
Tagging a new 2.x version will make sure PHP 5.3 users will keep using the current v2.6, while users of PHP >=5.6 will be using this version. This is not uncommon, ex: [DBAL](https://github.com/doctrine/dbal/releases) required PHP 7.3 when releasing a minor version (2.11.0). 
Thanks for your consideration.

pmjones:
> As far as I know it is not possible in the 2.x repo to support PHP 5.3 and 8.0 at the same time.

After some trials on my own, I see what you mean. Indeed, even though the test suite (with a class alias and an override of setExpectedException) can run on 5.3 all the way up to 7.4.x using PHPUnit 4 through 7, it cannot *also* run on PHP 8.0 and PHPUnit 8 or later, for various reasons (e.g. each() is removed in PHP 8.0, and the PHPUnit method TestCase signatures have changed with the addition of new typehints).

Even so, maintaining BC to PHP 5.3 *is* a design goal for this version of this package, and I cannot bring myself to break that BC promise. Yes, I know that some other projects opine that raising a PHP language version requirement is not a BC break, but [I (still) disagree with that opinion](https://paul-m-jones.com/post/2017/09/12/a-systems-addendum-to-semantic-versioning/).

I hate to deliver bad news, but all of the above means that (unless there's another way to get to it) I can't accept this PR -- it's not you, it's me.

One way forward, distasteful though you may find it, is for you to fork the Aura.Sql 2.x branch and modify it to suit your PHP 8.0 needs, until such time as you upgrade to Aura.Sql 3.x, or perhaps to one of the packages in the [Atlas](http://atlasphp.io) project.

My apologies for not being able to give you exactly what you want here.

pachico:
Thank you for your kind reply, @pmjones .
I do believe too having compatibility across many versions is a plus and should be kept.
However, you wouldn't offer less compatibility if you created a new tag with support for PHP 8, it would just be another version, which, as @sudo-plz says, it's quite a common practice.

I would hate to have to fork the project, to be honest.

pmjones:
> you wouldn't offer less compatibility if you created a new tag with support for PHP 8

That's true; according to the existing Aura rules, though, that would likely mean a 4.x branch, started from the 3.x branch, which is already incompatible with 2.x, which doesn't get you what you need. 

... Hm, unless you mean a one-off 2.x branch, not intended as part of the "normal" 2.x versioning? Something specifically to soothe PHP 8 and PHPUnit requirements, e.g., `2.x-PHP8`, that would not be part of the normal development process. Is that what you mean?

pachico:
> That's true; according to the existing Aura rules, though, that would likely mean a 4.x branch, started from the 3.x branch, which is already incompatible with 2.x, which doesn't get you what you need.

I see what you mean. I wasn't aware of incompatibilities with branch 2. In my case, I'm willing to pay the price for to accommodate changes due to upgrade, but I need something to upgrade to, which will never be the case if there's not support for PHP-8.
I guess, sooner or later this will be a commonly requested issue. I'd hate to have to either fork or move to another project.
Have you at least considered a family of projects which are more modern PHP friendly? Sort of Aura PHP 8 rules?

gravataLonga:
Any solution? maybe create a new branch only for this case? Composer allow require package by hashcommit, branch and tag. So, maybe can be a solution?

pachico:
@gravataLonga Not a branch, please. Remember that you aim to have reproducible builds and branches do not satisfy that.
I'd go for branch+tags to anchor your dependencies to.

pmjones:
@pachico  --

> I need something to upgrade to

I hear you. To sum up: you need something that works with *the code you have in place now* AND *works with PHP 8*. So no code changes, just a PHP upgrade.  You don't want to deal with fork, and for my part, I don't want to break previous promises, or get into the business of maintaining an PHP 8 version of the 2.x library.

> I'd hate to have to ... move to another project.

I get that; making code changes is a burden. Even so, the move to 3.x may not be *that* bad; see [here](https://github.com/auraphp/Aura.Sql/blob/3.x/docs/upgrade.md) for some upgrade notes, though of course it's set to PHP 5.6 and not 8.x. Switching to [Atlas.Pdo](http://atlasphp.io/cassini/pdo/) gets you up to PHP 7.x, which I expect will be closer to your PHP 8.x expectations, but it's a bigger code change than from Aura.Sql 2.x to 3.x.

Tell you what: I could start a 2.x branch that is **NOT** in the normal line of development (maybe call it `2.6.0-PHP8-x` or something else obvious), apply this PR to that branch, then tag it as `2.6.0-PHP8-a`, and never ever make more changes to it. That will give you some more space to choose what your longer-term upgrade path will be, whether to Aura.Sql 3.x, to Atlas.Pdo, or to something else entirely.

Thoughts?

pachico:
@pmjones thanks a lot for your proposal. I can clearly see that you are willing to take the required steps to help me and all those wanting to upgrade to 8.

> So no code changes, just a PHP upgrade.

Not really, I'm willing to do the required changes to upgrade. But, as I said, I need something to upgrade to and I strongly believe the community will ask for this sooner or later in a more frequent fashion.

Ideally, I must confess, I would like you to embrace PHP 8 as a target for your applications.
I suspect it's not possible to offer 5.4 to 8 compatibility for everything and there will be a moment in which all maintainers will face certain decisions to make.
So, if you create a version that is compatible to all versions out there, I'm happy to change what's required and stick to your package.

I can also help writing/testing.

pmjones:
> Ideally, I must confess, I would like you to embrace PHP 8 as a target for your applications.

Understandable! But for database work, I confess in return that I'm focusing my efforts more at Atlas and less at Aura these days; take a look at [Atlas.Pdo](http://atlasphp.io/cassini/pdo/) and let me know if that will suit your needs. When I start on the path to PHP 8, for me, it will be there rather than here.

Does that begin to help?

gravataLonga:
@pmjones like that, to me it's completely ok.

pmjones:
For what it's worth, I have started the Atlas.Pdo 2.x series, supporting PHP 8, here: https://github.com/atlasphp/Atlas.Pdo/tree/2.x

gravataLonga:
@pmjones i've not interested in Atlas.Pdo. Sorry, it's seems a good library but on my side its to much work to change to another library. If can't provide PHP8.0 support on this library will use another library but not Atlas.Pdo.. Sorry.

pachico:
@pmjones thanks for your message.
Unfortunately, if I was forced to change library I'm afraid I would use another one or I would do a fork that would allow me not to have to do any work to adapt my software to it.
Again, thanks for your time.

brandonsavage:
Unfortunately, dropping PHP 5.3 compatibility is not an option for this version of the package. Version 3.x supports 5.6 and better, and `master` runs against PHP 8.0. We recommend using this package with 3.x if you need to run PHP 8. 

As always, we're open to pull requests that honor the spirit of backwards compatibility for different versions, and hope that you'll continue to use Aura in your future applications.

PR #191, opened 2021-10-05
allow psr/log v2

r4ndsen:
widens the restriction for which logger interface to use.
similar to how symfony handles this: https://github.com/symfony/console/blob/5.3/composer.json#L34

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
     ],
     "require": {
         "php": ">=5.6.0",
-        "psr/log": "^1|^2"
+        "psr/log": "^1.0 || ^2.0"
     },
     "autoload": {
         "psr-4": {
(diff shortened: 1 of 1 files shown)


r4ndsen:
can you please review (and merge) this?

composer.json
```diff
@@ -23,7 +23,7 @@
     ],
     "require": {
         "php": ">=5.6.0",
-        "psr/log": "^1.0"
+        "psr/log": "^1|^2"
```
kenjis:
```suggestion
        "psr/log": "^1.0 || ^2.0"
```
https://getcomposer.org/doc/articles/versions.md#version-range

r4ndsen: the two expressions are equivalent, however i adjusted them to match your suggestion

kenjis: @r4ndsen Thank you!

PR #202, opened 2021-11-13
Enable PHP 7.2-8.1 compat

iansltx:


koriym (approved):

src/AbstractExtendedPdo.php
```diff
@@ -634,20 +636,18 @@ public function prepareWithValues(string $statement, array $values = []): PDOSta
...
-     * @param mixed ...$fetch_mode_args Optional fetch-related parameters.
+     * @param mixed ...$fetch Optional fetch-related parameters.
      *
      * @return PDOStatement
      *
      * @see http://php.net/manual/en/pdo.query.php
      *
      */
-    public function query(string $query, ?int $fetchMode = null, mixed ...$fetch_mode_args): PDOStatement
+    public function query(string $query, ...$fetch): PDOStatement
```
kenjis, line 646: This is different from 5.x.
But it seems 3.x is the same.
https://github.com/auraphp/Aura.Sql/blob/f19b1aa810b35d86ad55562876d967ef3f9a97b7/src/AbstractExtendedPdo.php#L645

Is this okay?

iansltx: I had to revert to that to avoid type signature conflicts in either 8.1 or earlier versions. Specifically, the default fetch mode per docs is a constant that evaluates to 19, but that constant is out of range for PDO itself rather than evaluating to whatever the default on-create waa. By contrast, defaulting to null and passing that through failed type checks as well. The only thing that worked across the board was the 3.x version.

If you swap the signature for the 5.x version, or the 5.x version without typehints that are only available in 8.0+, you'll see the errors that I saw this morning :)

kenjis: Okay, I understand 4.x should have the method signature.

src/AbstractExtendedPdo.php
```diff
@@ -564,18 +566,18 @@ public function perform(string $statement, array $values = []): PDOStatement
...
+     * @param array|null $options Set these attributes on the returned
      * PDOStatement.
      *
      * @return PDOStatement
      *
      * @see http://php.net/manual/en/pdo.prepare.php
      *
      */
-    public function prepare(string $query, array $options = []): PDOStatement
+    public function prepare($query, $options = null): PDOStatement
```
kenjis, line 577: 3.x
https://github.com/auraphp/Aura.Sql/blob/f19b1aa810b35d86ad55562876d967ef3f9a97b7/src/AbstractExtendedPdo.php#L577

The `$options` default value is `[]` on 3.x.
Why do you need to change to `null`?

iansltx: I think this was a signature issue with PDO, but I'll double check.

src/AbstractExtendedPdo.php
```diff
@@ -205,7 +205,8 @@ public function errorInfo(): array
      * @see http://php.net/manual/en/pdo.exec.php
      *
      */
-    public function exec(string $statement): int
+    #[\ReturnTypeWillChange]
```
srjlewis, line 208: Is #[\ReturnTypeWillChange] needed within the 4.x branch, since #[\ReturnTypeWillChange] is needed for PHP >= 8.1 to hide depreciation notices.

src/AbstractExtendedPdo.php
iansltx: Yep. Confirmed that that was the exact issue, see https://github.com/auraphp/Aura.Sql/runs/4199471301?check_suite_focus=true.

Just force-pushed again to revert that change to maintain <8.0 compat.

iansltx:
Bumping this as I believe all feedback has been responded to.

src/AbstractExtendedPdo.php
kenjis: It seems `$options` default value is `[]`.
>  public PDO::prepare(string $query, array $options = []): PDOStatement|false
https://www.php.net/manual/en/pdo.prepare.php#refsect1-pdo.prepare-description

Can you please explain it should be null?

iansltx: Sorry, thought I had posted the comment above. Hopefully that clears things up.

kenjis:
On PHP 7.3 and 7.4
```
There was 1 error:

1) Aura\Sql\ConnectionLocatorTest::testNullDefault
Declaration of Aura\Sql\AbstractExtendedPdo::prepare($query, array $options = Array): PDOStatement should be compatible with PDO::prepare($statement, $options = NULL)

/home/runner/work/Aura.Sql/Aura.Sql/src/AbstractExtendedPdo.php:26
/home/runner/work/Aura.Sql/Aura.Sql/src/ExtendedPdo.php:22
/home/runner/work/Aura.Sql/Aura.Sql/tests/ConnectionLocatorTest.php:24
```

src/AbstractExtendedPdo.php
kenjis: Thank you!

kenjis (approved):

Issue #183, opened 2019-06-26
Bug in Queryparser for SQLite


r4ndsen:
in sqlite a columnname containing `:` is valid, however the queryparser currently does not support this.

adjusting the `$skip` and `$split` fixed it for us

PR: https://github.com/auraphp/Aura.Sql/compare/3.x...r4ndsen:patch-2

r4ndsen:
```php
$query = 'select "a:A", \'B:b\', `c:c`, :d from table', 
$bind = ['d' => 'foo'];
```
^ this how works, previously it was trying to bind the non-existing `:c` param

r4ndsen:
as you can see in the example 
`"a:A"` is a string and is treated as such
`'B:b'` is a string and is treated as such
`c:c` refers to the column `c:c` directly and should not be part of the prepared statement. however this is where the bug in the query parser hits.

by adding backtick to the "specially treated characters" this problem is avoided

harikt:
This has been fixed. Closing for now.

PR #203, opened 2021-12-09
Update SqliteParser.php

r4ndsen:
as describes here: https://github.com/auraphp/Aura.Sql/issues/183

in sqlite a columnname containing `:` is valid, however the sqlite queryparser currently does not support this.

adjusting the `$skip` and `$split` fixed it for us

### DIFF
--- a/src/Parser/SqliteParser.php
+++ b/src/Parser/SqliteParser.php
@@ -17,4 +17,20 @@ namespace Aura\Sql\Parser;
  */
 class SqliteParser extends AbstractParser
 {
+    /**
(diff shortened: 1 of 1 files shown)


harikt:
We need a test so there will be no regression. Is it possible?

harikt:
Hi @r4ndsen,

Do you want this go to 3.x only or 4.x and 5.x as well ? Also if possible write a test case.

harikt:
I have added a test. @koriym I guess we can merge this feature.

And may be release 3.x soon.

PR #204, opened 2022-01-12
5.x - fix broken return types - adjust minimum php version to 8.0 since no 8.1 features are being used

r4ndsen:


Morthy:
It would be great to see this merged because this is currently preventing us from upgrading to PHP 8.1 in our project

harikt:
@Morthy what feature in this PR is preventing you to upgrade to PHP 8.1 ?

Looks all tests in 8.1 is working https://github.com/auraphp/Aura.Sql/runs/4577260986?check_suite_focus=true

composer.json
```diff
@@ -22,7 +22,7 @@
         }
     ],
     "require": {
-        "php": ">=8.1",
+        "php": ">=8.0",
```
harikt: I don't think the project will switch to 8.0. You may use 4.x in that case.

Or is there any issue using 4.x ?

Morthy:
> @Morthy what feature in this PR is preventing you to upgrade to PHP 8.1 ?
> 
> Looks all tests in 8.1 is working https://github.com/auraphp/Aura.Sql/runs/4577260986?check_suite_focus=true

This PR fixes the fatal TypeErrors that occur due to the wrong return types present in the 4.x and 5.x branch. If the 3.x branch is used, there are instead many deprecation notices due to incompatibility with the PDO methods. So basically 4.x and above are unusable on PHP 8.1 if you have any case like doing a fetchObject() call that returns no rows.

harikt:
@Morthy ok, I will look into this.

composer.json
r4ndsen: i reduced the requirement because the project runs perfectly fine on php 8.0 and does not need to be restricted to `>8.1`

iansltx: The discussion around this happened elsewhere and it was determined that Aura 5.x was 8.1+ only. Even though the codebase works fine on 8.0 now, maintainers wanted to give themselves the option to add 8.1+ syntax later without breaking compat with anyone who has 5.x already installed. Which is why I built out 4.x as a bridge up to 8.1.

harikt: I agree with @iansltx . If @r4ndsen can revert the changes for 8.0 , I will merge ASAP and make a release. Else it may take some time than expected.

r4ndsen: i adjusted my `5.x` branch - this pr

src/PdoInterface.php
```diff
@@ -186,5 +186,5 @@ public function setAttribute(int $attribute, mixed $value): bool;
      * @return array
      *
      */
-    public static function getAvailableDrivers(): array;
+    public static function getAvailableDrivers();
```
harikt: I missed this. Any reason why return type is removed ?

I am looking at https://www.php.net/manual/en/pdo.getavailabledrivers.php .

`public static PDO::getAvailableDrivers(): array`

composer.json
harikt: Cool. Thank you.

harikt:
Note : In the mean time if you want to push your code to 8.1 you can also do something like in example https://getcomposer.org/doc/05-repositories.md#loading-a-package-from-a-vcs-repository .

src/PdoInterface.php
r4ndsen: running unittests on `php8.0` reveals this:
```
Fatal error: Declaration of PDO::getAvailableDrivers() must be compatible with Aura\Sql\PdoInterface::getAvailableDrivers(): array in Aura.Sql/src/AbstractExtendedPdo.php on line 0
```

the typehint is fine when running on `php8.1`

i guess i changed it to be compatible with `php8.0`

// reverting it

harikt:
Hi all, 

Thanks for your contributions.

I have tagged 5.0.1 .

src/PdoInterface.php
sudo-plz: You can stay 8.0 compatible by adding the following to AbstractExtendedPdo.php since its fine to be more strict in child classes.
```
    // Overwrite to support PHP 8.0
    public static function getAvailableDrivers(): array
    {
        return PDO::getAvailableDrivers();
    }
```

Issue #194, opened 2021-11-01
Type hint incomplete


afilina:
The type hint suggest that it can only be an `array`, but if we dig into the [method](https://www.php.net/manual/en/pdostatement.fetch.php#refsect1-pdostatement.fetch-returnvalues) whose result we return, we see that it returns `false` on failure. This causes problems with static analysis tools when we then attempt to compare the result to `false`.

https://github.com/auraphp/Aura.Sql/blob/2754f95d99f995ea9f8e6b27d3f670e133ee465d/src/AbstractExtendedPdo.php#L407

Example usage:
```php
$row = $db->fetchOne($sql);

if ($row === false) {
  // handle failure
}
```

r4ndsen:
👍 
the same is true for https://github.com/auraphp/Aura.Sql/blob/4.x/src/AbstractExtendedPdo.php#L348

the method `fetchObject(..): object` does in fact *not* only return `object`. it returns `object|false`
i made it working with this hack:

```php
try {
  return $pdo->fetchObject(...)
} catch (\TypeError) {
  return false;
}
```

please note that the issue is also present in the `5.x` branch

r4ndsen:
we have provided this pull request
https://github.com/auraphp/Aura.Sql/pull/204

harikt:
Thank you @r4ndsen for your contribution. This has been fixed in 5.x branch and released 5.0.1 hope that helps.

I will check if this can be fixed in 4.x also.

harikt:
@afilina It looks like we will not be able to fix this in 4.x to support 8.1 features.

The union type is not supported in 7.x . So leaving as it is. In case if you have some other fix I will be happy to in-corporate those. Can we close this as we have 5.0.1 support 8.x ?

Thank you.

afilina:
@harikt I don't have a fix, so feel free to close it if it's addressed in a later version.

harikt:
Hi guys, 

I made a PR #209 so that 3.x will continue to work on 8.1. Please have a look into it, and report any issues.

I was mainly trying to address the YOURLS issue by @LeoColomb .

The doc comments may not be addressed correctly. So if anyone is willing to send a PR I will merge it. The rest of the code hopefully should work without any issues.

If merged I will make a new release and close this issue as well.

Thank you

harikt:
This has been addressed and 3.1.0 is released. Thank you guys for your support.

PR #208, opened 2022-02-23
Fix return types

LeoColomb:
Closes https://github.com/auraphp/Aura.Sql/issues/194


harikt:
Hi @LeoColomb ,

Can we make this change for 3.x ? May be also add #[ReturnTypeWillChange] attribute.

Issue #177, opened 2018-09-06
Bug with pgsql query


chekalsky:
If you will try to do this:

`$pdo->fetchCol("SELECT id FROM table WHERE removed = false AND data @> '{\"is_hidden\":false}'::jsonb AND type in (:types)", ['types' => [1, 2]]);`

you will get an error from DB: `SQLSTATE[08P01]: <<Unknown error>>: 7 ERROR:  bind message supplies 0 parameters, but prepared statement "pdo_stmt_00000001" requires 1`

And this is because `:types` will be skipped [here](https://github.com/auraphp/Aura.Sql/blob/3.x/src/Parser/AbstractParser.php#L154).

chekalsky:
Anybody?

chekalsky:
@pmjones is Aura.SQL abandoned?

pmjones:
Which version are you trying this with: v2, or v3?

chekalsky:
[v3](https://github.com/auraphp/Aura.Sql/blob/3.x/src/Parser/AbstractParser.php#L154)

harikt:
@chekalsky sorry for being late here. I actually added a failing test.

The fun fact is if you change the order 

```sql
SELECT * FROM table WHERE type in (:types) AND removed = false AND data @> '{\"is_hidden\":false}'::jsonb
``` 

this will work.

I don't have a fix right now. But can look into it at some point of time.

PR #216, opened 2022-02-28
Add failing test for #177

harikt:
This one is interesting when you change the order it will work.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -10,6 +10,8 @@ jobs:
 
     runs-on: ubuntu-latest
 
+    timeout-minutes: 10
+
     strategy:
       matrix:
         operating-system:
--- a/tests/Parser/AbstractParserTest.php
+++ b/tests/Parser/AbstractParserTest.php
@@ -148,4 +148,24 @@ SQL;
         $this->assertEquals($statement, $sql);
         $this->assertEquals($values, $parameters);
     }
+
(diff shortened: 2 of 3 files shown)


kenjis:
@harikt What do you want? 

The test fails, because `"::jsonb AND type in (:types)"` goes into the `if` statement.

https://github.com/auraphp/Aura.Sql/blob/a6d9fa87009659340a632e1f2ef365e7b329fe11/src/Parser/AbstractParser.php#L154-L156

kenjis:
https://github.com/auraphp/Aura.Sql/blob/a6d9fa87009659340a632e1f2ef365e7b329fe11/src/Parser/AbstractParser.php#L308

Now
```
SELECT * FROM table WHERE type in (:types) AND removed = false AND data @> '{\"is_hidden\":false}'::jsonb
```
will be split into
```
0 = "SELECT * FROM table WHERE type in (:types) AND removed = false AND data @> "
1 = "'{\"is_hidden\":false}'"
2 = "::jsonb"
```
But should be
```
0 = "SELECT * FROM table WHERE type in (:types) AND removed = false AND data @> "
1 = "'{\"is_hidden\":false}'::jsonb"
```

harikt:
@kenjis Your findings are very interesting.

What I am trying to show here is the test is failing. So we need to fix it. I didn't got the time to look into it.

If you find some time to fix it, please do send a PR.

I think this will be same for 5.x also.

PR #218, opened 2022-08-03
Enable PHP 8.2 compat

koriym:
There are no changes in the source code. The current release `3.1.0` also works fine with PHP 8.2. This PR is just a fix for a problem in the test code in PHP8.2.

@see https://wiki.php.net/rfc/deprecate_dynamic_properties

### DIFF
--- a/tests/FakeObject.php
+++ b/tests/FakeObject.php
@@ -1,7 +1,9 @@
 <?php
 namespace Aura\Sql;
 
-class FakeObject
+use stdClass;
+
+class FakeObject extends stdClass
 {
     public $foo;
 
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -16,6 +16,7 @@ jobs:
           - ubuntu-latest
         php-version:
           - '8.1'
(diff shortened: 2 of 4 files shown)


tests/FakeObject.php
```diff
@@ -1,7 +1,9 @@
 <?php
 namespace Aura\Sql;
 
-class FakeObject
+use stdClass;
```
harikt, line 4: Not sure why you want to extend stdClass.

harikt (commented):
Apart from that looks good to me.

harikt:
Can you send to 3.x also?

tests/FakeObject.php
koriym: Extending stdClass avoids the "Deprecate dynamic properties" error.

https://wiki.php.net/rfc/deprecate_dynamic_properties

harikt: But in this case we don't have dynamic properties right ? We already have defined the property. Did I missed something ?

koriym: Confirmed. We missed the id and name properties there.

PR #220, opened 2022-08-04
Enable PHP 8.2 compat (v3)

koriym:


tests/FakeObject.php
```diff
@@ -3,6 +3,8 @@
 
 class FakeObject
```
harikt, line 4: Did you missed to extend stdClass in this case. Or we are all good with just the 2 properties.

koriym: Adding properties fix the issue. see https://github.com/auraphp/Aura.Sql/pull/219

harikt: I noticed that PR after my comment. Cool.

Issue #213, opened 2022-02-28
Abandon 4.x ?


harikt:
Hi all,

We have 3.x which supports from 5.6 to 8.1 now.

4.x currently have some issues due to return types as mentioned in #194 .

5.x is using > 8.1 which supports union types, so #194 is not a problem.

Considering the return type failure ( multiple return types not handled ) for 4.x I think we should drop support for 4.x .

Thoughts ?

iansltx:
As the person who built out 4.x to bridge between 3.x and 8.1, yeah, nuke it.

harikt:
Thank you @iansltx . I was waiting for others reply. But seems there is no objections. So we are officially abandoning 4.x .

PR #221, opened 2022-08-04
Enable PHP 8.2 compat (v4)

koriym:


harikt:
We have abandoned 4.x . See https://github.com/auraphp/Aura.Sql/issues/213 .

PR #223, opened 2022-09-20
Allow using psr/log v3

Qrzysio:
Why not using the third version of popular `psr/log` dependency?

`"psr/log": "^1.0 || ^2.0 || ^3.0",`


harikt:
Hi there, 

I have not been taking very close look into the psr changes. ie why.

Thank you for the PR and it looks good to me.

composer.json
```diff
@@ -23,7 +23,7 @@
     ],
     "require": {
         "php": ">=8.1",
-        "psr/log": "^1.0 || ^2.0",
+        "psr/log": "^1.0 || ^2.0|| ^3.0",
```
harikt: One thing I wished is if there is a space between `2.0` and `|` . Not a big deal just pointing.

koriym:
```suggestion
        "psr/log": "^1.0 || ^2.0　|| ^3.0",
```

harikt:
Thank you.

harikt:
Interesting the test failed after merge.

composer.json
```diff
@@ -23,7 +23,7 @@
     ],
     "require": {
         "php": ">=8.1",
-        "psr/log": "^1.0 || ^2.0",
+        "psr/log": "^1.0 || ^2.0　|| ^3.0",
```
kenjis, line 26: Should be `"^1.0 || ^2.0 || ^3.0"`.
There is a zero width space (`　`) after `2.0`.

PR #229, opened 2024-10-03
Changes for PHP 8.4

srjlewis:


harikt:
This can go to 6.x , else we need to make sure the connect method is not removed or another alternate approach needed .

src/ExtendedPdo.php
```diff
@@ -100,7 +113,11 @@ public function connect(): void
         // connect
         $this->profiler->start(__FUNCTION__);
         list($dsn, $username, $password, $options, $queries) = $this->args;
-        $this->pdo = new PDO($dsn, $username, $password, $options);
+        if(version_compare(phpversion(), '8.4.0', '<')) {
```
frederikbosch: I'd remove this check and only support PHP 8.4

.github/workflows/continuous-integration.yml
```diff
@@ -18,6 +18,7 @@ jobs:
           - '8.1'
           - '8.2'
           - '8.3'
+          - '8.4'
```
frederikbosch, line 18: I'd drop all older versions.

src/ExtendedPdo.php
```diff
@@ -85,13 +85,26 @@ public function __construct(
...
+    public static function connect(
+        string $dsn,
+        ?string $username = null,
+        ?string $password = null,
+        ?array $options = [],
+        array $queries = [],
+        ?ProfilerInterface $profiler = null
+    ): static {
+        $pdo = new static($dsn, $username, $password, $options ?? [], $queries, $profiler);
+        $pdo->establishConnection();
```
frederikbosch: Do we really want to connect immediately when using driver specific constructor? I'd say we always connect lazily.

frederikbosch (commented):
Two discussions points: which versions to support and should the static function connect return an immediate connection?

composer.json
```diff
@@ -22,7 +22,7 @@
         }
     ],
     "require": {
-        "php": ">=8.1",
+        "php": ">=8.4",
```
frederikbosch: While I am using ~8.4.0 for all my packages, I know others more like `^8.4.0`. I would definitely not use `>= 8.4` because that would allow PHP 9 to also install the upcoming version while we don't know if we support PHP 9. But I don't know if I am alone on this one or not.

frederikbosch (commented):
Single discussion left

composer.json
srjlewis: I update as is, but I think it is personaly preferance I think, but PHP could break it between minor versions as well, like 8.3 to 8.4, so I dont think it would make much difference

frederikbosch: Would you agree to go for `^8.4`? Then we don't have to do anything for PHP 8.5, but PHP 9 requires at least changing the composer.json.

srjlewis: Yes I think thats resonable

frederikbosch:
@harikt and @koriym, move this one in?

frederikbosch:
Thanks for the effort @srjlewis. I might have advocated my points strongly but I really appreciate the work.

harikt:
I am ok with the changes. Thank you guys.

PR #230, opened 2024-11-15
Add establishConnection() for interoperability with future versions

srjlewis:
This allows eaiser switch between v5 and v4 of Aura.Sql and different PHP versions

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -32,7 +32,7 @@ If you want to force a connection, call the `establishConnection()` method.
 
 > Previous `connect()` method has been deprecated due to the introduction of 
 > ```PDO::connect()``` in [PHP 8.4](https://www.php.net/releases/8.4/en.php#pdo_driver_specific_subclasses), 
-> so we encourage users tp use `establishConnection()` instead.
+> so we encourage users t0 use `establishConnection()` instead.
 
 ```php
 // does not connect to the database
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -30,8 +30,9 @@ actually needs the connection to the database; e.g., on `query()`.
 
 If you want to force a connection, call the `establishConnection()` method.
 
-> Previous `connect()` method has been deprecated and we encourage users move 
-> to using `establishConnection()`
+> Previous `connect()` method has been deprecated due to the introduction of 
+> ```PDO::connect()``` in [PHP 8.4](https://www.php.net/releases/8.4/en.php#pdo_driver_specific_subclasses), 
+> so we encourage users tp use `establishConnection()` instead.
 
 ```php
 // does not connect to the database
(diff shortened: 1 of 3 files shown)


harikt:
Can you update the docs also ? Thank you.

srjlewis:
I would think adding to the upgrade notes about the change, and update the getting started section, sound about right.

Also would it be a good idea to backport this to 3.x and above versions or just keep it to 5.x?

We also would need to update 6.x docs aswell with the new ```connect()``` method

srjlewis:
I have update the docs, but please any other suggestions to the docs would be welcome.

docs/upgrade.md
```diff
@@ -1,3 +1,44 @@
+# 5.x Upgrade Notes
```
harikt, line 1: 6.x right ?

docs/getting-started.md
```diff
@@ -28,7 +28,10 @@ Whereas the native _PDO_ connects on instantiation, _ExtendedPdo_ does not
 connect immediately. Instead, it connects only when you call a method that
 actually needs the connection to the database; e.g., on `query()`.
 
-If you want to force a connection, call the `connect()` method.
+If you want to force a connection, call the `establishConnection()` method.
+
+> Previous `connect()` method has been deprecated and we encourage users move
```
harikt: May be we link to the php doc also ?

> Previous `connect()` method has been deprecated due to introduction of xxx in ..... So we encourage users to use `establishConnection()` instead.

Thoughts ?

docs/upgrade.md
srjlewis: No I meen 5.x as it is a deprecation in 5.x and adding a alias for ```connect()``` method, as the commit is targeting 5.x branch.

I need to make PR for the doc changes in 6.x which explain the new connect method.

docs/getting-started.md
srjlewis: I have updated the docs to be a bit clearer

docs/getting-started.md
```diff
@@ -28,7 +28,11 @@ Whereas the native _PDO_ connects on instantiation, _ExtendedPdo_ does not
 connect immediately. Instead, it connects only when you call a method that
 actually needs the connection to the database; e.g., on `query()`.
 
-If you want to force a connection, call the `connect()` method.
+If you want to force a connection, call the `establishConnection()` method.
+
+> Previous `connect()` method has been deprecated due to the introduction of 
+> ```PDO::connect()``` in [PHP 8.4](https://www.php.net/releases/8.4/en.php#pdo_driver_specific_subclasses), 
+> so we encourage users tp use `establishConnection()` instead.
```
harikt: tp => typo .

srjlewis: I will hit the right key at some point. should be fixed now

docs/getting-started.md
```diff
@@ -32,7 +32,7 @@ If you want to force a connection, call the `establishConnection()` method.
 
 > Previous `connect()` method has been deprecated due to the introduction of 
 > ```PDO::connect()``` in [PHP 8.4](https://www.php.net/releases/8.4/en.php#pdo_driver_specific_subclasses), 
-> so we encourage users tp use `establishConnection()` instead.
+> so we encourage users t0 use `establishConnection()` instead.
```
harikt: oh you still making the same error. `to` . is this 0 ?

PR #240, opened 2025-04-18
Update .scrutinizer.yml

harikt:


harikt:
@koriym can we remove the scrutinizer ?

PR #245, opened 2025-05-24
Reintroduce PHP 8.1-8.3 support

### LINKED ISSUE #238: Add PHP 8.4 support to v5
I just got updated to PHP 8.4 on my dev machine by my OS package manager, and to my surprise my project stopped working with `PHP Fatal error:  Cannot make static method PDO::connect() non static in class Aura\Sql\AbstractExtendedPdo`.

I realize that v6 of this library was release for PHP 8.4 support, which is great, but the project I'm using this in _must_ have cross-version compatibility (8.2+ at the moment), so it's a big problem for the library to not work across versions. It'll also making rolling out the PHP version update _a lot_ trickier because we'll have to sync deploying a code change at the same time as the PHP server upgrade, adding complexity to the flow.

In v5, could we do a `php_version()` condition and have two versions of the `AbstractExtendedPdo` class, with the `connect()` function removed from the PHP 8.4 version? Our codebase doesn't use `connect()` at all anyway, we use the `new` constructor approach to creating connections and we always just lazy connect.

If not that, could v6 be adjusted to retain 8.1+ support, with just `connect()` dropped compared to v5?

francislavoie:
Closes https://github.com/auraphp/Aura.Sql/issues/238

Pretty straight forwards. Makes version-specific copies of `ExtendedPdo` and `PdoInterface` so that they continue to work on older PHP versions.

The `ExtendedConnectPdoTest` test is skipped since `::connect()` isn't a thing for earlier PHP versions.

I had to disable the PHPUnit `processUncoveredFiles` coverage option because it would scan these version-specific files and throw `PHP Fatal error:  Cannot declare class Aura\Sql\ExtendedPdo`.

I adjusted the heredocs in the tests to use the PHP 7.3+ indented heredoc syntax, it was just bothering me to look at :sweat_smile: 

The difference between the two files are as such:

```
$ diff src/ExtendedPdoPHP83.php src/ExtendedPdoPHP84.php                             
88,90c88,94
<     public function connect(): void
<     {
<         throw new \RuntimeException('The connect() method is no longer supported in Aura.Sql v6. Use lazyConnect() instead.');
---
>     public static function connect(
>         string $dsn,
>         ?string $username = null,
>         ?string $password = null,
>         ?array $options = []
>     ): static {
>         return new static($dsn, $username, $password, $options);


$ diff src/PdoInterfacePHP83.php src/PdoInterfacePHP84.php                     
46a47,69
>      * Introduced in 6.x due to PHP 8.4 change. This is a BC break for Aura.Sql.
>      *
>      * @param string $dsn The Data Source Name, or DSN, contains the information required to connect to the database.
>      *
>      * @param string | null $username  The user name for the DSN string. This parameter is optional for some PDO drivers.
>      *
>      * @param string | null $password The password for the DSN string. This parameter is optional for some PDO drivers.
>      *
>      * @param array | null $options  A key=>value array of driver-specific connection options.
>      *
>      * @return \PDO Returns an instance of a generic PDO instance.
>      *
>      * @see https://www.php.net/manual/en/pdo.connect.php
>      */
>     public static function connect(
>         string $dsn,
>         ?string $username = null,
>         #[\SensitiveParameter] ?string $password = null,
>         ?array $options = null
>     ): static;
> 
>     /**
>      *
```

In english: `PdoInterface` has the `::connect()` method dropped entirely, and `ExtendedPdo` adds a non-static `connect(): void` which throws an exception warning that it can no longer be used as such in earlier PHP versions (was deprecated in v5 anyway).

## Summary by CodeRabbit

- **New Features**
  - Expanded support for PHP versions 8.1, 8.2, and 8.3, in addition to 8.4+.
  - Added new changelog entry documenting recent changes and compatibility notes.

- **Bug Fixes**
  - Corrected class name case in test exception handling.

- **Refactor**
  - Modularized code to provide separate implementations for different PHP versions.
  - Updated interface and class structures to align with PHP version-specific features.

- **Style**
  - Improved code formatting and indentation in test files for better readability.

- **Chores**
  - Updated continuous integration to test across multiple PHP versions.
  - Adjusted PHPUnit configuration to prevent errors with uncovered files.
  - Broadened PHP version constraint in package requirements.
  - Enhanced sensitive data handling by marking connection parameters as sensitive.


francislavoie:
Scrutinizer is complaining, but that's not new, looks like it has been failing for a while on the 6.x branch.

francislavoie:
Ah interesting, I broke something with trying to fix one of the complaints from the AI bot. Will fix.

harikt:
I don't know how well we can support this. The interface is changing based on php version and that is actually a breaking change according to semversion. That is why I have not replied very quickly. I am eager to hear from others what they think.

francislavoie:
It's not a breaking change because it's introduced to v6 only which already had PHP 8.1-8.3 support removed anyway. It's purely an addition of new functionality, which is _not_ a break. Nothing changes at all for people on PHP 8.4+.

koriym:
@harikt @francislavoie 

I'm also hesitant about this approach for the same semver concerns @harikt mentioned.

Additionally, the conditional file loading pattern adds complexity that hasn't been part of Aura.Sql's design philosophy.

(Sorry for the delayed response.)

francislavoie:
This is not really "added complexity". This is a very common pattern which is used by Symfony polyfills for example which is used by almost everyone.

But anyway, because I didn't want to wait anymore, I copied my fork into our project as a package replacement since we needed it right away. So I'm not blocked by this not being merged anymore. But it's still quite annoying that proper care wasn't taken to keep PHP version compatibility between Aura versions when it's not that difficult to keep.

koriym:
Thanks for understanding and for implementing your own solution. Good luck with your project!

PR #252, opened 2026-09-18
Mark $password as SensitiveParameter

koriym:
## Problem

`PdoInterface::connect()` declares `#[\SensitiveParameter]` on `$password`, but the implementations do not. When a connection fails, `DecoratedPdo::connect()` keeps the plaintext password as a frame argument in the exception backtrace.

Reproduced on `6.x` (PHP 8.4), comparing against native PDO in the same script:

| `zend.exception_ignore_args` | before | after |
| --- | --- | --- |
| `0` (CLI default, `php.ini-development`) | **password present in `Aura\Sql\DecoratedPdo::connect` frame args** | clean |
| `1` (`php.ini-production`) | clean (arguments are not recorded at all) | clean |

Native `PDO::__construct` is clean under both, because PHP marks its `$password` sensitive. Only the wrapper passed it through.

`ExtendedPdo::__debugInfo()` already redacts the credentials for `var_dump()`, so this is the remaining hole in the same intent: "The purpose of this method is to hide sensitive data from stack traces."

## Change

Add `#[\SensitiveParameter]` to the `$password` parameter of:

- `DecoratedPdo::connect()`
- `ExtendedPdo::__construct()`
- `ExtendedPdo::connect()`

This aligns the implementations with the interface that already declares it. No behaviour change beyond backtrace redaction, no BC break, no new dependency. `composer.json` requires `php: ^8.4`, so the attribute needs no version guard.

Note that `#[\SensitiveParameter]` is not inherited: a subclass overriding `connect()` or `__construct()` without repeating the attribute will expose the password in its own frame again. That is a PHP limitation rather than something this change can address.

## Test

`DecoratedPdoTest::testConnectKeepsPasswordOutOfStackTrace` triggers a connection failure with an unreachable sqlite DSN (no network, consistent with the rest of the suite), then scans `$e->getTrace()` for the password.

Two details keep the test from passing vacuously:

- It sets `zend.exception_ignore_args=0` itself (restoring the old value in `finally`). Without this the test is green even with the bug when the suite runs under production ini settings.
- It asserts that the non-sensitive sibling argument (`$username`) *is* present in the recorded args, proving arguments were captured and that the redaction is selective.

Verified:

- without the fix: `Tests: 1, Assertions: 6, Failures: 1`
- with the fix: `OK (1 test, 6 assertions)`
- full suite before: `191 tests, 434 assertions, 3 errors`; after: `192 tests, 440 assertions, 3 errors`. The 3 errors are pre-existing on an unmodified tree (`testCall` / `sqliteCreateFunction`) and unrelated to this change.

## Not covered here

`var_export($pdo)` still exposes the plaintext, because it ignores `__debugInfo()` and dumps the protected `$args`. Closing that would mean not retaining the plaintext in `$args` at all, which is a larger design change than this PR.

## Summary by CodeRabbit

* **Security**
  * Passwords are now automatically redacted from stack traces and error logs when database connections fail.

* **Tests**
  * Added coverage verifying that database passwords do not appear in exception stack traces.


harikt:
Thank you

